PR #93, opened 2021-04-15
Add github action to run integration tests on push

mattkasun:


mattkasun:
this should have been to develop not master

PR #178, opened 2021-06-09
Allow for the SERVER_HOST to be v6 and tokens to still work

karmanyaahm:
Use net package to concatenate hostnames


karmanyaahm:
I just reset it on a new machine and there seems to be something wrong with netclient, I'll investigate.

karmanyaahm:
netclient also needed the same functions to be changed and now everything works AFAICT

PR #198, opened 2021-07-16
Feature docs

JonathanBouligny:
The install command appears to no longer be in the netlcient so I removed a reference to it, also i fixed two lines in the docs like we previously discussed.


JonathanBouligny:
fixing branching

Issue #221, opened 2021-08-11
Netclient fails when interface does not exist.


hedgty:
If the nm-\<network name\> interface does not exist, `netclient checkin -n <network name>` will fail and output `failed to parse interface`. This is a situation that occurs after reboots. The network will work again after running `netclient pull -n <network name>`.

I was able to temporarily fix this issue by making the changes in the diff file to the code and rebuilding netclient.
http://hastebin.com/xukovozego.diff

afeiszli:
fixed in the latest hotfix. Thanks for the help @hedgty!

PR #222, opened 2021-08-11
Fix issue with netclient failing when interface doesnt exist, issue #221

hedgty:
I implemented a way to fix issue #221.


afeiszli:
Looks great! Thank you for your contribution and quick fix to this issue, we will merge this and make a hotfix today.

PR #228, opened 2021-08-12
Hotfix 0.7.1 add additional peers

dillon-carns:
We weren't telling clients to fill in the allowed ips.

Just had them fetch it whenever they pull peers.


afeiszli (approved):
Looks good

PR #345, opened 2021-10-13
feat(models): increase node name length

restanrm:
Custom server name can have more than 12 chars.


restanrm:
Thank you ! I looked into the code and didn't find limitations. If there is any issue with some distro, don't mind this PR.

Issue #349, opened 2021-10-14
Use HTTPS for ip checks


kusold:
HTTPS should be preferred over HTTP.

https://github.com/gravitl/netmaker/blob/6b5332c6e070293a1ef5c6478db332de9fe8ad3c/netclient/ncutils/netclientutils.go#L92

I validated that https://api.ipify.org and https://ipinfo.io/ip support HTTPS. I'd like to request http://ip.client.gravitl.com support it as well.

afeiszli:
added in 0.9.0

PR #350, opened 2021-10-14
[#349] Use HTTPS for ip checks

kusold:
This uses HTTPS for all external IP Checkers that support it.


afeiszli:
@kusold Thanks for the PR. The purpose of the CLA is because we've been debating switching to a more open license down the road, but we will have problems doing that if we don't have pre-approval from all contributors. However, I understand the concern.

Issue #352, opened 2021-10-15
New machine get random name


restanrm:
When a machine join a network for the first time it gets a random name. This step requires a manual action later to fix the name of the machine (that could be hard if multiple servers join the network at the same time). I would be easier to set the name of the machine that join the network to it's hostname.

PR #353, opened 2021-10-15
Feat/change default hostname

restanrm:
This PR proposes a solution for the issue #352 

The server may want to refuse the join operation if the hostname already exits in the network.


0xdcarns:
Tested, nice easy change

PR #419, opened 2021-10-31
check if files exists, if not get from latest

pete1019:


bsherman:
Looks like this will address the concern I had. 👍

PR #425, opened 2021-11-03
armv7l fix

pete1019:
Pi Zero 2 or some Pi 3 will have armv7l. They need armv7.
uname -m
armv7l

it printed a 404 error because thre is no armv7l file


afeiszli:
merging, thanks!

PR #437, opened 2021-11-04
Docker compose remove host binding for Caddy

Belphemur:
Hi,

It's better to avoid using `network_mode: host` as much as possible.
If you need it for binding privliedge port like `80` and `443` then it's better to use `capabilities: cap_net_bind_service`.

I'm currently running this config without any issue, caddy is listening on the right ports and we can remove unneeded port bindings by relying on the links

### DIFF
--- a/compose/docker-compose.caddy.yml
+++ b/compose/docker-compose.caddy.yml
@@ -16,6 +16,7 @@ services:
       - NET_ADMIN
       - SYS_ADMIN
     restart: always
+    network_mode: host
     privileged: true
     environment:
       SERVER_HOST: "SERVER_PUBLIC_IP"
--- a/compose/docker-compose.caddy.yml
+++ b/compose/docker-compose.caddy.yml
@@ -43,8 +43,6 @@ services:
     image: gravitl/netmaker-ui:v0.8.5
     links:
       - "netmaker:api"
-    ports:
-      - "8082:80"
     environment:
       BACKEND_URL: "https://api.NETMAKER_BASE_DOMAIN"
     restart: always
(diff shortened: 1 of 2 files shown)


afeiszli:
Hi @Belphemur, to run netmaker with host networking off, you can use the "contained" compose file: https://github.com/gravitl/netmaker/blob/master/compose/docker-compose.contained.yml

This was actually the main compose file until we hotfixed 0.8.5 last night. Reason being: a lot of people are using the server as a bastion for testing connections and SSH'ing to other machines. This is not feasible without host networking on.

Belphemur:
@afeiszli Hooo I see. That is definitely a usecase for it (I was actually in the middle of installer teleport on a node for that reason).

I'll update the PR to reenable the host network for netmaker. The rest still apply though, Caddy doesn't need it :)

Belphemur:
There is one issue with the current version of the docker-compose file, the network devices created by netmaker don't get deleted when the container goes down. So when it goes back up, it complain that the port is already in use.

afeiszli:
Can you open a ticket for that interface issue and we will take a look? I am aware of it but it will draw attention so we can prioritize. I think all we need to do is just run a delete/recreate if it already exists.

afeiszli:
These changes look good, thank you.

Belphemur:
@afeiszli No problem.

issue created: https://github.com/gravitl/netmaker/issues/440

afeiszli:
@Belphemur I just reverted these changes in develop. The changes to caddyfile + docker-compose cause a CORS issue, and the UI cannot connect with the backend. Will review later but don't want to have them in develop until we figure out a fix.

Issue #457, opened 2021-11-14
Docker netmaker:test-workflow arm64 wrong binary


youpko:
I tried running the Netmaker server on a raspberry pi but found that the ARM version of the server is not supported (yet).
But reading through the issues I tried following the advice given in issue #447 to use the `test-workflow` tag in the docker compose file.
- netmaker-ui : start up normally now 
- netmaker : crashes with `standard_init_linux.go:228: exec user process caused: exec format error`. 

After pulling the netmaker binary form the docker image and inspecting it i see that is a x86-64 binary
```
root@ubuntu:~# file netmaker
netmaker: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, Go BuildID=DozHLjGzQXN1CX8d5AXs/Ij5aXYbnEvf40M_WlYir/2kjyLDII_i15xrENb0-_/LX510dk1sKb4J09-nwoI, stripped
```
I double checked that I pulled the arm64 image.
```
gravitl/netmaker:test-workflow
Digest:sha256:4e74dbc472614a96e13638d652cc7509ce0e4f22e0d0fcb11a5f06a34e1c0f13
```

afeiszli:
@mattkasun can we add netmaker binary to the release workflow please?

mattkasun:
PR #633 adds netmaker binary to release workflow

PR #480, opened 2021-11-17
Fix workflow Publish Docker arm64 images as well as x86

mattkasun:
switched to golang-alpine for builder in Dockerfile
Changes workflow to run on release rather than pull request
added rudimentary test to make sure container runs before publishing


mattkasun:
fixes #457

PR #499, opened 2021-11-22
Update install script: netclient-armv7 to netclient-arm7

a1ad:
Wrong url generated by the script. The file to download is ...-arm7 and not ...-armv7


a1ad:
The FreeBSD entry probably also needs to change to arm7, but i don't have a BSD ARM machine to test.

PR #518, opened 2021-11-26
Added -y parameter

wnark:
update will not take effect without adding "-y"


afeiszli:
upgrade commands require "-y", but update commands do not

PR #532, opened 2021-12-03
Fix getting tag in Docker Publish workflow

desty2k:
Get branch name/release tag from `github.ref_name` instead of `github.base_ref`. "base_ref property is only available when the event that triggers a workflow run is either pull_request or pull_request_target."

Source: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context


mattkasun (approved):
This is good, we should have made this change when we switched from pull-requests to publish/workflow-dispatch as the workflow triggers.

mattkasun:

PR #541, opened 2021-12-09
Replace SystemD with systemd

frafra:
systemd is lowercase, as described in https://www.freedesktop.org/wiki/Software/systemd/


mattkasun:

PR #553, opened 2021-12-16
Update README.md

rajeshg007:
Bash should be started with sudo permissions not wget.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@
 2. Open ports 443, 80, 53, and 51821-51830/udp on the VM firewall and in cloud security settings.
 3. Run the script **(see below for optional configurations)**:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,7 +43,7 @@
 2. Open ports 443, 80, 53, and 51821-51830/udp on the VM firewall and in cloud security settings.
 3. Run the script **(see below for optional configurations)**:
 
-`sudo wget -qO - https://raw.githubusercontent.com/gravitl/netmaker/master/scripts/nm-quick.sh | bash`
```
rajeshg007, line 46: Sudo should be applied to bash command and not the wget command.
This was tested on ubuntu 20.04

mattkasun:

afeiszli (approved):
makes sense, we usually run as root but this will fix it for those running as non-root

PR #557, opened 2021-12-21
Replace SliceContains by StringSliceContains

tranngoclam:


mattkasun:
please update the target of this PR to gravitl:develop vice master

0xdcarns:
As of 0.9.2, I believe we have this function elsewhere, but we should use it where this is being called.

PR #562, opened 2021-12-26
(draft): makefile to build multiple target

oycol:
I found new way to build multiple target, it 's so simple, but it still have some bug with it, i can't  release it on my repository ，so maybe some one  can handle it. I am a salt fish 🐟 .

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,35 @@
+name: Release
+on: [ push ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Cache go module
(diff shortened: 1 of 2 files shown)


oycol:
with make -j command , you can build every  target  on make file . example : make -j linux-amd64 ,it will run command on the  linux-amd64 line.

oycol:
source from [clash](https://github.com/Dreamacro/clash) repo

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
+name: Release
+on: [ push ]
```
mattkasun, line 2: Why run this on every push?

mattkasun (changes requested):
The current buildandrelease workflows builds multiple clients when a release is published or it can be manually dispatched.  What advantage does this have over the existing workflow.

.github/workflows/release.yml
oycol: It's just one build  way，I test it on my repo so,  I only provide a way to build,  with makefile to build rather than release flow , everything can be changed, I have no idea with the release flow ，so I set it as draft

PR #593, opened 2022-01-07
typo fix

pr0PM:
Fixed a minor typo in database section.


pr0PM:
then we can close this right? it will rise up here from dev?

PR #607, opened 2022-01-15
wireguard/unix: add checks for key validity `StorePrivKey`

ok-john:
Encountered an issue when trying netmaker out for the first time, `StorePrivKey` was failing due to the parent directory returned from `ncutils.GetNetclientPathSpecific()` not existing.

Added two checks in `StorePrivKey`:
1. Check to ensure the passed `key` is a valid WireGuard private key.
2. Check the parent directory where `key` is stored exists/create it if not.

Additionally switched the permissive bits on the private key file from `0644` to `077` to ensure that only the user creating the key file will have read/write access.

`StorePrivKey` currently & prior to this PR will overwrite existing keys, perhaps that's desired functionality, perhaps not.

Added tests, since the path used to store private keys is under `/etc/..` the tests will fail unless executed by a privileged user. 

In general, seeing a private key passed to a public function as a string is concerning to me, but out of scope for this simple PR.


afeiszli:
@ok-john thank you for the PR! This is definitely a good security update. My main note is that the function is meant to be able to overwrite an existing key, so that keys can be cycled if necessary.

I also agree it makes sense to put it in a private function. It's only referenced in 2 places on the client so we can probably move it.

One last note is that PR's should be against develop, but that's on us for not providing clear contribution guidelines.

afeiszli (changes requested):
@ok-john thank you for the PR! This is definitely a good security update. My main note is that the function is meant to be able to overwrite an existing key, so that keys can be cycled if necessary.

I also agree it makes sense to put it in a private function. It's only referenced in 2 places on the client so we can probably move it.

One last note is that PR's should be against develop, but that's on us for not providing clear contribution guidelines.

ok-john:
> My main note is that the function is meant to be able to overwrite an existing key, so that keys can be cycled if necessary.

For sure, will remove the comment code mentioning it.

> I also agree it makes sense to put it in a private function. It's only referenced in 2 places on the client so we can probably move it.

I went ahead and switched it to private, however, some pieces of code out of this functions scope use it. ([example](https://github.com/gravitl/netmaker/blob/master/netclient/functions/join.go#L185)). Could switch the scope around by adjusting the project layout to have `storePrivateKey` within scope or by leaving `StorePrivateKey` public/exported. Thoughts? 

> One last note is that PR's should be against develop, but that's on us for not providing clear contribution guidelines.

Ah my bad, switched the PR to develop.

ok-john:
@afeiszli FWIW my concerns, which are ultimately opinions (and should be treated as such), are mainly with private key material being passed to a function that is creating a deep copy with dynamically allocated memory and less about the export scope of the function. Here's [another example](https://github.com/gravitl/netmaker/blob/9602d6a5f1fd4fb9077c91f87599bfd6bf384f66/logic/wireguard.go#L313) where you're creating a dynamically allocated copy in memory that isn't from passing it to a function by-value. 

While unavoidable at times, private keys should exist as infrequently as possible in memory and ideally once you're done using said material are ensuring it's discarded appropriately. This isn't straight-forward to do in Go at the moment but there is an [open issue with under implementation](https://github.com/golang/go/issues/21865) to allow for overriding the GC for this exact reason. When that is implemented, and should you decide this to be relevant to your application, you'll be having to chase down the multiple deep copies of the same key that exist in memory which may be even more difficult.

Like I said though, I think this is out of the scope of this PR, I just wanted express that I don't see the public export of the function as a concern in comparison to the additional copies of key material.

PR #615, opened 2022-01-17
Feature v0.10.0 UUID switch

0xdcarns:
- Converted clients/nodes to use a UUID rather than be mac address dependent.


afeiszli (approved):
Lots of changes but generally straightforward stuff that makes sense; adds in parameters to functions for nodeID and switches out apis to use nodeID. Will definitely require a lot of testing before release but code looks good.

PR #601, opened 2022-01-11
add Cli parameter

louis-zhou:
resolved #600

PR #625, opened 2022-01-19
add cli parameter

louis-zhou:
V0.9.4 #601


afeiszli (changes requested):
@louis-zhou Please sign the CLA, then this will be good to merge

louis-zhou:
> 
I finally found the reason for not signing the CLA. I used the wrong email address . I will resubmit it.

PR #649, opened 2022-01-26
Feature v0.10.0 server response

0xdcarns:
- refactored join
- refactored getting server ips from server for MQ connections

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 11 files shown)


logger/logger.go
```diff
@@ -36,6 +37,9 @@ func ResetLogs() {
 
 // Log - handles adding logs
 func Log(verbosity int, message ...string) {
+	var mu sync.Mutex
```
afeiszli, line 40: What is the mutex for?

0xdcarns: The logger is being accessed by 3 different go routines in v0.10.0 and this helps avoid concurrent map writes. Tested and it doesn't affect performance

0xdcarns:
rebased on latest dev changes

logger/logger.go
0xdcarns: has been removed

PR #728, opened 2022-02-10
update Caddyfile to fix CORS

chaychoong:
Following the Quick Install instructions will result in the dashboard encountering CORS issues.

This PR fixes that by enabling Access-Control-Allow-Origin on the Caddyfile


ElectronicElephant:
I think there is still some problem here...

image

PR #740, opened 2022-02-11
Feature v0.10.1 strings

0xdcarns:


logger/util.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"os"
+	"strconv"
+	"strings"
+)
+
+// MakeString - makes a string using golang string builder
+func MakeString(delimeter string, message ...string) string {
+	var builder strings.Builder
+	for i := 0; i < len(message); i++ {
```
mattkasun, line 12: why not
 `for i := range message {`

PR #839, opened 2022-02-25
move wg to container

mattkasun:
- remove bind mount to /usr/bin/wg from docker-compose files
- added wireguard-tools to netmaker container (Dockerfiles)


mattkasun:
> As a bystander, I'm genuinely curious why `wg` hasn't been included with the docker image.

legacy artifact from early version where netclient was run on the host server rather than inside the netmaker container

0xdcarns (approved):

PR #861, opened 2022-03-04
Set ext_client mtu to match ingress gw

### LINKED ISSUE #857: MTU mismatch between ingress/egress gateway and external clients
The default MTU when setting up an ingress + external clients is 1280 for the ingress and nothing for the generated external client configurations, which means they use the WireGuard default which is 1420.

This has mostly worked fine but it has caused intermittent issues. What is the idea behind the 1280 default? And why are the external clients not made to match it? Is it a mistake or is there a reasoning behind it?

patstrom:
Currently the external client configuration does not specify a MTU,
which causes WireGuard to use the default of 1420. The default of the
ingress gateway will be 1280. This mismatch can cause network issues.

This patch ensures that the generated external client configuration is
specified to use a MTU that matches that one the ingress gateway.

I have to admit that I have not been able to test this, since I don't really
have a testing setup in the first place. Although it seems like this should
do the trick.

Fix #857

### DIFF
--- a/controllers/ext_client.go
+++ b/controllers/ext_client.go
@@ -154,6 +154,7 @@ func getExtClientConf(w http.ResponseWriter, r *http.Request) {
 	config := fmt.Sprintf(`[Interface]
 Address = %s
 PrivateKey = %s
+MTU = %d
 %s
 
 [Peer]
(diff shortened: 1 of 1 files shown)


afeiszli:
Hi @patstrom, thanks for this PR! Can you please sign the CLA, and switch the branch to be against develop (it's currently open against master) then we can merge

0xdcarns:
CLA wasn't signed in time, we added something similar on your behalf, thanks for the find.

patstrom:
Yeah sorry about that I had to confirm with my employers, and they still haven't got back to me.

PR #880, opened 2022-03-08
don't update dns on comms network peer update

mattkasun:


0xdcarns:
Don't think this is a "feature" branch lol

PR #905, opened 2022-03-15
added mutex for sqlite

0xdcarns:


database/sqlite.go
```diff
@@ -60,6 +61,9 @@ func sqliteCreateTable(tableName string) error {
 
 func sqliteInsert(key string, value string, tableName string) error {
 	if key != "" && value != "" && IsJSONString(value) {
+		var mutex sync.Mutex
+		mutex.Lock()
```
mattkasun, line 65: var mutex sync.Mutex should be a package variable

mattkasun (changes requested):

PR #919, opened 2022-03-18
netclient: pass by value -> reference

ok-john:
Updates various function arguments to accept `config.ClientConfig`
by reference instead of by value.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,7 +6,6 @@ package config
 
 import (
 	"fmt"
-	"log"
 	"os"
 
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -110,7 +109,6 @@ func readConfig() (*EnvironmentConfig, error) {
 
 func init() {
 	if Config, SetupErr = readConfig(); SetupErr != nil {
-		log.Fatal(SetupErr)
-		os.Exit(2)
+		Config = &EnvironmentConfig{}
```
mattkasun, line 112: This should have been a separate PR rather than piggybacking on to a larger PR

PR #922, opened 2022-03-20
server routes

mattkasun:
updating server routes uses same logic as updating routes of peers.


mattkasun:
Integration test failure is expected due to PR#882

Issue #930, opened 2022-03-21
[Bug]:no bugs.I just can't find a 32-bit netclient for windows


perfectstorm01:
### What happened?

I can't find a 32-bit netclient for windows, would you consider making one, I found that it can't be used for 32-bit windows during deployment, thank you very much for considering my problem, this work is really amazing, thank you again for being able to use it.

### Version

v0.12.0

### What OS are you using?

Windows

### Contributing guidelines

- [ ] Yes, I did.

perfectstorm01:
In addition, I would like to know whether to consider the automatic routing function of windows, that egress function always prompts that it cannot be configured on windows

ok-john:
Hey @perfectstorm01, this will build a 32-bit netclinet.
```bash
$ GOARCH=386 GOOS=windows GOHOSTARCH=amd64 CGO_ENABLED=0 go build -o bin/windows/netclient-386
```
I'll add these to our artifacts in the coming releases.

perfectstorm01:
I appreciate your reply and thoughtfulness，but I'm not very good at programming, could you help me compile a 32-bit netclient, and then hang it in the attachment, or send it to my email：perfectstrom01@163.com. Grateful!

> Hey @perfectstorm01, this will build a 32-bit netclinet.
> 
> ```shell
> $ GOARCH=386 GOOS=windows GOHOSTARCH=amd64 CGO_ENABLED=0 go build -o bin/windows/netclient-386
> ```
> 
> I'll add these to our artifacts in the coming releases.

I appreciate your reply and thoughtfulness，but I'm not very good at programming, could you help me compile a 32-bit netclient, and then hang it in the attachment, or send it to my email：perfectstrom01@163.com. Grateful!

si458:
the latest 0.12.2 doesnt work on 32bit Windows 10 either

afeiszli:
32bit windows is not supported. If users need 32-bit support, they will need to compile themselves as listed above.

PizzaProgram:
Can somebody please tell me:
- where can I get the latest version of netclient_x86.msi or exe for **Win7 32bit** ?

It is not fair that it's:
- nowhere written, that `netclient.exe` is only for 64 bit.
_(Not at readme.md, nor at docs)_

Also `netclient_x86.msi` is suggesting it can run on both OSes !
The setup process does not show any error message, it is finishing as if everything would be alright... 👎 
[...]

PR #948, opened 2022-03-22
scripts/build-binaries.sh: support for all OS/ARCH

ok-john:
Expands our build scripts to support the following OS/ARCHS:
```
netclient-linux-amd64
netclient-linux-arm5
netclient-linux-arm6
netclient-linux-arm7
netclient-linux-arm64
netclient-linux-mips
netclient-linux-mips64
netclient-linux-mips64le
netclient-linux-mipsle
netclient-linux-ppc64
netclient-linux-ppc64le
netclient-linux-riscv64
netclient-linux-s390x
netclient-linux-386
netclient-freebsd-amd64
netclient-freebsd-arm5
netclient-freebsd-arm6
netclient-freebsd-arm7
netclient-freebsd-arm64
netclient-freebsd-386
netclient-darwin-arm64
netclient-darwin-amd64
netclient-windows-amd64
netclient-windows-arm5
netclient-windows-arm6
netclient-windows-arm7
netclient-windows-arm64
netclient-windows-386
```

### DIFF
--- a/netclient/bin-maker.sh
+++ b/netclient/bin-maker.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-if [[ ${#VERSION} -lt 1 ]]; then VERSION="develop"; fi 
+VERSION=${VERSION:-"develop"}
 echo "build with version tag: $VERSION"
 readonly __HOST_ARCH=${1:-"amd64"}  # change this for your machine.
 readonly __HOST_GOOSE=${2:-"linux"} # change this for your machine.
--- a/netclient/bin-maker.sh
+++ b/netclient/bin-maker.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+if [[ ${#VERSION} -lt 1 ]]; then VERSION="develop"; fi 
+echo "build with version tag: $VERSION"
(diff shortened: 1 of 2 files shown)


ok-john:
closes #930

netclient/bin-maker.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+if [[ ${#VERSION} -lt 1 ]]; then VERSION="develop"; fi
```
mattkasun, line 2: replace lines 1-2 with 
echo "build with version tag '${VERSION:-develop}'"

mattkasun (changes requested):

netclient/bin-maker.sh
ok-john: That doesn't update the value of VERSION in the case where it is not set.

ok-john: Checks if unset, setting it otherwise. 
```bash
if [[ ${#VERSION} -lt 1 ]]; then VERSION="develop"; fi  
```
relevant for..
```bash
-ldflags="-X 'main.version=$VERSION'"
```

mattkasun (approved):

PR #955, opened 2022-03-22
config/version: switch VERSION to constant

ok-john:
Switches Version from a runtime var to a constant.
This constant, VERSION, should be updated manually
with each semantic version change & commit to master.

The format is:

	semvar-commithash[:7]

For example:

	v0.12.1-c784f99

Hardcoding the semantic version as a constant allows us to implement
controls for preventing mismatched net{client,maker}
binaries from interacting.

These controls become critical to the security of netmaker networks
in the event of a vulnerability wherein the resolving patch increments
the semnatic version. In this scenario, controls to automatically
prevent vulernable (read: out-of-date) binaries from interacting
with the latest, patched binaries are crucial.

### DIFF
--- a/.github/workflows/buildandrelease.yml
+++ b/.github/workflows/buildandrelease.yml
@@ -18,11 +18,7 @@ jobs:
     steps:
       - name: Get Version Number
         run: |
-          if [[ -n "${{ github.event.inputs.version }}" ]]; then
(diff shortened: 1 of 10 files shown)


0xdcarns (commented):
You will have to discuss with @mattkasun as this will affect the workflows

PR #960, opened 2022-03-24
reworked usage/description for netclient and added verbosity flags

0xdcarns:


netclient/cli_options/flags.go
```diff
@@ -204,5 +204,23 @@ func GetFlags(hostname string) []cli.Flag {
 			Value:   "no",
 			Usage:   "Allows to run the command with force, if otherwise prevented.",
 		},
+		&cli.BoolFlag{
+			Name:    "verbosity-level-1",
+			Aliases: []string{"V"},
```
mattkasun, line 209: upper case?  isn't lower case more common.

netclient/config/config.go
```diff
@@ -25,7 +25,6 @@ type ClientConfig struct {
 	Daemon          string         `yaml:"daemon"`
 	OperatingSystem string         `yaml:"operatingsystem"`
 	DebugOn         bool           `yaml:"debugon"`
```
mattkasun, line 27: This var can be deleted and rework where it is used to log by verbosity to achieve the same functionality

mattkasun (commented):
couple of questions.  Changes not necessarily required.

netclient/cli_options/flags.go
0xdcarns: yeah you're right, I'll change to lowercase, I was running into issues with lowercase at some point and left at uppercase for some reason

PR #965, opened 2022-03-25
fix: Fixed wrong Image Tag

Hugo54x:
Updated image tag to `gravitl/netmaker-ui:v0.12.1`


afeiszli:
Thank you for your contribution, but we are changing the versions in develop to v0.12.2, otherwise we would merge this

PR #985, opened 2022-03-29
Fix docker config to avoid exposing ports unnecessarily

fersingb:
Hello,

I gave netmaker a try today and it looks great. However the "quick install" method using docker-compose seems to expose ports to the outside world unnecessarily. Even though this could be "fixed" by setting firewall rules I think it's better to not expose those ports in a first place since it makes the "quick install" more secure by default.

This PR is an attempt to fix this issue by using the network created by docker-compose to communicate between the containers instead of exposing all the services to 0.0.0.0 on the host.

I tested the fix with the nm-quick.sh script and everything looks good, however I'm not 100% sure my changes don't have an impact on other deployment methods (especially the Caddyfile changes). Let me know what you think about it.

Cheers,
Boris

### DIFF
--- a/compose/docker-compose.contained.yml
+++ b/compose/docker-compose.contained.yml
@@ -39,8 +39,6 @@ services:
       VERBOSITY: "1"
     ports:
       - "51821-51830:51821-51830/udp"
-      - "8081:8081"
-      - "50051:50051"
   netmaker-ui:
     container_name: netmaker-ui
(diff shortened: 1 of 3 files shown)


fersingb:
I used the wrong email address to commit the changes and couldn't accept the CLA, I opened a new PR with the correct email address. Sorry for the noise.

PR #1015, opened 2022-04-13
Feature v0.13.0 mq tls

mattkasun:
adds tls to mosquitto connections.

nm-quick script updated to generate server certs.  the root.pem certificate has to be manually copied into /etc/netclient/<server name>/ on each client in order for the client to be able to successfully connect.


mattkasun:
superseded by #1016

PR #1024, opened 2022-04-22
Feature v0.13.0 dockerfiles

mattkasun:
-  create a builder image which includes all go mod dependencies so that these dependencies don't have to be downloaded every time
- results in ~40% time reduction to build netmaker
- includes a workflow to update the builder image once a week (Sun evening)

### DIFF
--- a/.github/workflows/docker-builder.yml
+++ b/.github/workflows/docker-builder.yml
@@ -3,7 +3,7 @@ name: Build go-builder images
 on:
   workflow_dispatch:
   schedule:
-    -cron: '00 21 * * SUN'
+    - cron: '00 21 * * SUN'
 
 jobs:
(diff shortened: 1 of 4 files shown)


0xdcarns (approved):

MohamedAmineOuali:
Hello,

I believe this change made it impossible to build images for the arm64 architecture when using `golang:1.18.0-alpine3.15` I am able to build the image form arm64. 

In the docker hub website  `gravitl/go-builder` is only built for linux/amd64.

I also tried the 0.13.0 netclient image it no longer works on arm64.

Best regards.

mattkasun:
> 

PR #1051 now builds gravitl/go-builder for arm64 as well

PR #1071, opened 2022-05-08
Use TCP instead of ICMP for MQTT check-in

whites11:
When trying to connect azure VMs using Nat Gateways to a netmaker network, the MQTT ping feature fails because ICMP is not supported on Azure (see https://docs.microsoft.com/en-us/azure/virtual-network/nat-gateway/troubleshoot-nat#how-to-validate-connectivity). Note that the same thing might happen in other networking-restricted environments as well.

This PR switched from using ICMP to a plain TCP ping against the broker address and port.

The reasoning behind this is that, while ICMP might be blocked, as long as TCP connection towards the broker port is up'n'running we should consider the client alive.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 
 require (
 	filippo.io/edwards25519 v1.0.0-rc.1
-	github.com/go-ping/ping v0.0.0-20211130115550-779d1e919534
+	github.com/cloverstd/tcping v0.1.1
 	github.com/guumaster/hostctl v1.1.2
 	github.com/kr/pretty v0.3.0
 	github.com/posthog/posthog-go v0.0.0-20211028072449-93c17c49e2b0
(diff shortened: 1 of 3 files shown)


afeiszli:
Thanks for this PR! However, we have already removed the ping in develop, because it was not necessary. @mattkasun do you think this could be helpful to have in the codebase?

whites11:
> Thanks for this PR! However, we have already removed the ping in develop, because it was not necessary. @mattkasun do you think this could be helpful to have in the codebase?

Oh ok, good to know.
I tested the latest release (0.13) yesterday and faced this issue, but happy to wait for next release and get this removed.

TBH I was questioning the need for ping when there is an open tcp connection to the same endpoint.

afeiszli:
Yeah it was necessary with a previous design but is redundant now. I will close this for now, but let us know if you encounter a similar issue in the next week. And please do consider contributing in the future! This is good code regardless.

mattkasun:
> Thanks for this PR! However, we have already removed the ping in develop, because it was not necessary. @mattkasun do you think this could be helpful to have in the codebase?

This looks like a very useful package and I can see uses for it in the future

PR #1103, opened 2022-05-19
add checks/logs on broker connect timeout

mattkasun:
using code suggested by @whites11 in PR #1071


netclient/functions/daemon.go
```diff
@@ -242,6 +242,14 @@ func setupMQTT(cfg *config.ClientConfig, publish bool) mqtt.Client {
 		} else {
 			err = token.Error()
 		}
+		brokerErr := checkBroker(cfg.Server.Server)
+		if brokerErr.Error() == "dns" {
```
0xdcarns, line 246: this code block 246-252 is repeated below, so probably want to convert to a private func and replace it. You also don't check if that err is `nil` like you do below, which could cause a panic.

netclient/functions/mqpublish.go
```diff
@@ -140,7 +142,16 @@ func publish(nodeCfg *config.ClientConfig, dest string, msg []byte, qos byte) er
 			err = token.Error()
 		}
 		if err != nil {
-			return token.Error()
+			if brokerErr := checkBroker(nodeCfg.Server.Server); brokerErr != nil {
+				if brokerErr.Error() == "dns" {
```
0xdcarns, line 146: see above, turn lines 146-152 into private func.

0xdcarns (changes requested):

netclient/functions/daemon.go
mattkasun: moved logging to checkBroker func

netclient/functions/mqpublish.go
mattkasun: moved logging to checkBroker func

0xdcarns (approved):

PR #1115, opened 2022-05-23
removing legacy DNS logic

afeiszli:


Exchizz:
Hi,

You comment the following lines:
https://github.com/gravitl/netmaker/blob/master/netclient/wireguard/common.go#L342 

I would like to be able to push a DNS server my windows clients (managed by netclient) should use. Will these lines be included again or will there be a new way to push a DNS server to Windows clients ? 

Thanks in advance

PR #1148, opened 2022-05-30
changing broker retrieval/setting logic

afeiszli:


mattkasun (changes requested):
GUI changes are required with remove of Server from access token

afeiszli:
updated with GUI change

PR #1154, opened 2022-05-31
add docker-compose support for traefik instead of caddy

bsherman:
This adds a new docker-compose.traefik.yml file to use Traefik instead of Caddy.

It includes configuration which is ready for MQTT over TLS passthrough with Traefik on port 443, but not enabled as the current builds of Netmaker/Netclient do not support this yet.

### DIFF
--- a/compose/docker-compose.traefik.yml
+++ b/compose/docker-compose.traefik.yml
@@ -1,0 +1,141 @@
+version: "3.4"
+
+services:
+  netmaker:
+    container_name: netmaker
+    image: gravitl/netmaker:v0.14.1
+    cap_add: 
+      - NET_ADMIN
+      - NET_RAW
+      - SYS_MODULE
+    sysctls:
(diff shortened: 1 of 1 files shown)


compose/docker-compose.traefik.yml
```diff
@@ -0,0 +1,141 @@
...
+      - netmaker
+    restart: unless-stopped
+    volumes:
+      - /root/mosquitto.conf:/mosquitto/config/mosquitto.conf
+      - mosquitto_data:/mosquitto/data
+      - mosquitto_logs:/mosquitto/log
+      - shared_certs:/mosquitto/certs
+    ports:
+      - "127.0.0.1:1883:1883"
+    # comment once netmaker supports changing MQ port
```
mattkasun, line 123: This should be updated to for netmaker v0.14.2+  i.e. MQ port is configurable:

mattkasun (changes requested):

Issue #1184, opened 2022-06-04
[Bug]: windows /etc/hosts file repeating entries


si458:
### What happened?

when checking my `c:\windows\system32\drivers\etc\hosts` file on ANY of my windows machines
the file is neally 1MB in size, 
when opening it, it is just repeating entry and entry and entry of ALL my servers connected to my netmaker network?

### Version

v0.14.2

### What OS are you using?

Windows

### Relevant log output

```shell
# profile.on nm-my
10.28.38.4 maycocksserver.myNetwork
10.28.38.8 ColinServer.myNetwork
10.28.38.3 Win2016.myNetwork
10.28.38.254 netmaker-1.myNetwork
10.28.38.5 SalramServer.myNetwork
10.28.38.1 Win2012R2.myNetwork
# end

# profile.on nm-my
10.28.38.4 maycocksserver.myNetwork
10.28.38.8 ColinServer.myNetwork
10.28.38.3 Win2016.myNetwork
10.28.38.254 netmaker-1.myNetwork
10.28.38.5 SalramServer.myNetwork
10.28.38.1 Win2012R2.myNetwork
# end

# profile.on nm-my
10.28.38.4 maycocksserver.myNetwork
10.28.38.8 ColinServer.myNetwork
10.28.38.3 Win2016.myNetwork
10.28.38.254 netmaker-1.myNetwork
10.28.38.5 SalramServer.myNetwork
10.28.38.1 Win2012R2.myNetwork
# end
```

### Contributing guidelines

- [X] Yes, I did.

si458:
so i created a new network called 'othernetwork' and im not seeing this issue now...
so it appears to be a `C`apital letter issue with the network name,
a simple fix would be in the create network to only allow lowercase letters

si458:
just checked my linux server and wow they are twice as worse! the /etc/hosts file was 15MB in size! with over 250,000 lines in it
so just cleared the /etc/hosts files out, rejoined to the my `othernetwork` and i havent seen any replicate entries now

afeiszli:
@si458 just want to clarify, this issue is present on both Windows and Linux, and is the result of having a capital letter in the network name, correct?

si458:
> @si458 just want to clarify, this issue is present on both Windows and Linux, and is the result of having a capital letter in the network name, correct?

That is all correct 👌

PR #1188, opened 2022-06-06
force lowercase for /etc/hosts profile names

mattkasun:
to test:
create network with Upper case name
check profile in /etc/hosts is created with all lowercase


mattkasun:
> @mattkasun if we do this, does that mean limiting networks to lowercase on the server is unnecessary?

correct.   rules for network names do not have to change.  This ensures that the /etc/hosts profiles are all lowercase

0xdcarns (approved):

0xdcarns:
Actually I think we should limit networks, otherwise you'd have net, Net, nEt all going to the same profiles right?

mattkasun:
fixes #1184

0xdcarns (approved):

PR #1194, opened 2022-06-08
Bugfix v0.14.3 server acl check

afeiszli:
In our testing, this was preventing server peers from being added. However, it has not been a problem till this point and should already be affecting the server in production, so unsure of why this is not always an issue.

### DIFF
--- a/logic/server.go
+++ b/logic/server.go
@@ -215,7 +215,7 @@ func GetServerPeers(serverNode *models.Node) ([]wgtypes.PeerConfig, bool, []stri
 	if err != nil {
 		logger.Log(1, "could not fetch current ACL list, proceeding with all peers")
 	}
-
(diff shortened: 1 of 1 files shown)


logic/server.go
```diff
@@ -215,7 +215,7 @@ func GetServerPeers(serverNode *models.Node) ([]wgtypes.PeerConfig, bool, []stri
 	if err != nil {
 		logger.Log(1, "could not fetch current ACL list, proceeding with all peers")
 	}
-
+	git
```
0xdcarns, line 218: stray `git`

0xdcarns (changes requested):

afeiszli:
fixed

0xdcarns (approved):

PR #1196, opened 2022-06-09
updating node can wipe traffic keys

ygelfand:
When updating nodes via api (/api/nodes/{network id}/{macaddress}, PUT) if the original traffic keys aren't provided, they are wiped, leaving the server unable to decrypt mq messages from the node.

This retains previous keys regardless if new ones are supplied, as it didn't appear that updating these keys manually to new values is a supported workflow.


afeiszli:
great PR! Thanks for submitting

PR #1203, opened 2022-06-13
update relay when relayed address updated

mattkasun:


logic/nodes.go
```diff
@@ -681,3 +681,24 @@ func unsetHub(networkName string) error {
 	}
 	return nil
 }
+
+func FindRelay(node *models.Node) *models.Node {
```
0xdcarns, line 685: missing comment

controllers/node.go
```diff
@@ -758,3 +764,25 @@ func runForceServerUpdate(node *models.Node) {
 func isServer(node *models.Node) bool {
 	return node.IsServer == "yes"
 }
+
+func updateRelay(oldnode, newnode *models.Node) {
+	relay := logic.FindRelay(oldnode)
+	newrelay := relay
+	if oldnode.Address != newnode.Address {
```
0xdcarns, line 771: I'd recommend putting a comment `if oldnode.Address != newnode.Address { // check for address difference to...` for future reference

0xdcarns (changes requested):

0xdcarns (approved):

PR #1204, opened 2022-06-13
Feature v0.14.3 zombies

mattkasun:
- will add zombie nodes to quarantine list at startup.
- when node is added, checks for node with same macAddress on the same network and adds the one with oldest checkin time to the quaratine list.
- when node is deleted, removes from quarantine list if present
- periodically checks if zombie has not checked in for significant period of time and if so, deletes node
- 

Testing:

- create server and some nodes:
- stop server
- leave network on client
- restart server
- join same network on client
- wait for zombie to be delete (10 minutes)

### DIFF
--- a/logic/zombie.go
+++ b/logic/zombie.go
@@ -9,8 +9,10 @@ import (
 )
 
 const (
-	ZOMBIE_TIMEOUT     = 60 // timeout in seconds
-	ZOMBIE_DELETE_TIME = 10 // minutes
+	// ZOMBIE_TIMEOUT - timeout in seconds for checking zombie status
+	ZOMBIE_TIMEOUT = 60
+	// ZOMBIE_DELETE_TIME - timeout in minutes for zombie node deletion
+	ZOMBIE_DELETE_TIME = 10
 )
 
 var (
@@ -44,13 +46,17 @@ func ManageZombies(ctx context.Context) {
 			logger.Log(1, "adding", id, "to zombie quaratine list")
 			zombies = append(zombies, id)
 		case id := <-removeZombie:
+			found := false
 			for i, zombie := range zombies {
 				if zombie == id {
 					logger.Log(1, "removing zombie from quaratine list", zombie)
 					zombies = append(zombies[:i], zombies[i+1:]...)
+					found = true
 				}
 			}
-			logger.Log(3, "no zombies found")
+			if !found {
+				logger.Log(3, "no zombies found")
+			}
 		case <-time.After(time.Second * ZOMBIE_TIMEOUT):
 			for i, zombie := range zombies {
 				node, err := GetNodeByID(zombie)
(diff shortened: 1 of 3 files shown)


logic/zombie.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/gravitl/netmaker/logger"
+	"github.com/gravitl/netmaker/models"
+)
+
+const (
+	ZOMBIE_TIMEOUT     = 60 // timeout in seconds
```
0xdcarns, line 12: comments should be:
```
// ZOMBIE_TIMEOUT - timeout in seconds for zombies..
ZOMBIE_TIMEOUT = 60
```

logic/zombie.go
```diff
@@ -0,0 +1,99 @@
...
+			logger.Log(1, "adding", id, "to zombie quaratine list")
+			zombies = append(zombies, id)
+		case id := <-removeZombie:
+			for i, zombie := range zombies {
+				if zombie == id {
+					logger.Log(1, "removing zombie from quaratine list", zombie)
+					zombies = append(zombies[:i], zombies[i+1:]...)
+				}
+			}
+			logger.Log(3, "no zombies found")
```
0xdcarns, line 53: this comment occurs every time regardless of if there are zombies or not..

logic/zombie.go
```diff
@@ -0,0 +1,99 @@
...
+		case <-time.After(time.Second * ZOMBIE_TIMEOUT):
+			for i, zombie := range zombies {
+				node, err := GetNodeByID(zombie)
+				if err != nil {
+					logger.Log(1, "error retrieving zombie node", zombie, err.Error())
+					continue
+				}
+				if time.Since(time.Unix(node.LastCheckIn, 0)) > time.Minute*ZOMBIE_DELETE_TIME {
+					if err := DeleteNodeByID(&node, true); err != nil {
+						logger.Log(1, "error deleting zombie node", zombie, err.Error())
```
0xdcarns, line 63: should this continue and not run the logic below if a deletion failure occurs?

logic/zombie.go
```diff
@@ -0,0 +1,99 @@
...
+					}
+					logger.Log(1, "deleting zombie node", node.Name)
+					zombies = append(zombies[:i], zombies[i+1:]...)
+				}
+			}
+		}
+	}
+}
+
+//InitializeZombies - populates the zombie quarantine list (should be called from initialization)
```
0xdcarns, line 73: missing space at beginning of comment `// InitializeZombies`

logic/zombie.go
```diff
@@ -0,0 +1,99 @@
...
+func InitalizeZombies() {
+	nodes, err := GetAllNodes()
+	if err != nil {
+		logger.Log(1, "failed to retrieve nodes", err.Error())
+		return
+	}
+	for _, node := range nodes {
```
0xdcarns, line 80: would be preferred to use index to save memory if looping through node structs

0xdcarns (changes requested):

logic/zombie.go
mattkasun: need access to node data

0xdcarns (approved):

PR #1241, opened 2022-06-22
Bugfix 0.14.3 for netclient-install.sh on opnsense(FreeBSD)

calebgasser:
### Issues
I ran into issues when using the netclient-install.sh script on opnsense(FreeBSD) according to what the documentation states. [Here is a related issue.](https://github.com/gravitl/netmaker/issues/1149) These changes fix a couple issues;

### Fixes
1. `error decoding token` when running without providing the `KEY` to `netclient-install.sh`. One approach would have been to [update the documentation](https://docs.netmaker.org/netclient.html#freebsd) to inform users to provide the `KEY` same as they do `VERSION`. However, if you just want to install the netclient and not immediately join a network the install script will break with the aforementioned error since `KEY` is set to the string `nokey` and it will try to join a network regardless. I added a check to see if the `KEY` is set to `nokey` ( as it does when one is not provided ). If it's not set, do not try to run `netclient join`.
2. Moving the netclient executable to a directory present on `PATH`.
3. Install `wget` as a dependency as this isn't present by default.

This fix has been tested on a fresh install of `OPNsense 22.1.8_1-amd64` which is built on `FreeBSD 13.0-STABLE`

### Existing bug found, but not corrected.
When first joining a network after installing the netclient on opnesense (or freebsd) with `netclient join -t <token>` you will get an error as it tries to start the service twice. I haven't tracked down what is causing this just yet. It doesn't break anything as you still join the network just fine. There is just unneeded code running trying to start the service multiple times.

### DIFF
--- a/netclient/ncutils/netclientutils_freebsd.go
+++ b/netclient/ncutils/netclientutils_freebsd.go
@@ -20,7 +20,7 @@ func RunCmdFormatted(command string, printerr bool) (string, error) {
 	out, err := cmd.CombinedOutput()
 	if err != nil && printerr {
 		logger.Log(0, "error running command: ", command)
-		logger.Log(0, strings.TrimSuffix(string(err), "\n"))
+		logger.Log(0, strings.TrimSuffix(string(out), "\n"))
 	}
 	return string(out), err
 }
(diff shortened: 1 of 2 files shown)


afeiszli (approved):

afeiszli:
thank you for the PR!

PR #1244, opened 2022-06-23
remove check as node has been already deleted from database

mattkasun:


logic/wireguard.go
```diff
@@ -346,9 +346,6 @@ func setWGKeyConfig(node *models.Node) error {
 }
 
 func removeLocalServer(node *models.Node) error {
-	if !IsLocalServer(node) {
```
mattkasun, line 349: node is deleted from database prior to call to removeLocalServer and thus !IsLocalServer will always be true.

0xdcarns (approved):

PR #1247, opened 2022-06-24
Implement OpenID Connect as an OAuth Provider

capric98:
With additional `Env: OIDC_ISSUER` or `Config: oidcissuer`, now user can login netmaker with any OIDC Provider.


capric98:
let me check develop branch and create the request again T_T

PR #1259, opened 2022-06-28
update version numbers to 14.4

mattkasun:
Signed-off-by: Matthew R. Kasun <mkasun@nusak.ca>


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -31,7 +31,7 @@ body:
       label: Version
       description: What version are you running?
       options:
-        - v0.14.3      
+        - v0.14.4
```
afeiszli, line 34: this should add rather than replace

Issue #1269, opened 2022-06-29
[Bug]: Cannot connect to egress network behind relay


edwardchenchen:
### What happened?

1 When I use the following network, everything works fine
<img width="594" alt=" 2022-06-29 at 14 09 16" src="https://user-images.githubusercontent.com/6397192/176349798-e9063c76-fdcb-4e8b-baeb-54d0e6f04308.png">
<img width="1065" alt=" 2022-06-29 at 14 11 59" src="https://user-images.githubusercontent.com/6397192/176349946-bf469701-cef8-48f1-900c-0c93f2c3b2b6.png">

but when I switch to this network, egress behind relay, it wont work, 192.168.201.2 is raspberrypi
<img width="550" alt=" 2022-06-29 at 14 12 43" src="https://user-images.githubusercontent.com/6397192/176350076-f1157e85-4be6-42f4-b615-f7feb5549b95.png">

<img width="1046" alt=" 2022-06-29 at 14 13 37" src="https://user-images.githubusercontent.com/6397192/176350163-172f4f96-f3d1-48e8-86ef-c48b0a7d8a09.png">

I see this issue https://github.com/gravitl/netmaker/issues/517

it mentioned Egress is now enabled behind relay.

### Version

v0.14.4

### What OS are you using?

Linux, Mac

### Contributing guidelines

- [X] Yes, I did.

edwardchenchen:
```
root@raspberrypi:~# wg show
interface: nm-test-net
  public key: removed
  private key: (hidden)
  listening port: 44077

peer: removed
  endpoint: 45.77.removed:56840
  allowed ips: 192.168.201.1/32, 192.168.201.3/32, 192.168.201.254/32, 192.168.201.4/32, 192.168.201.5/32, 192.168.201.6/32, 192.168.1.0/24
  latest handshake: 18 seconds ago
  transfer: 96.46 KiB received, 112.66 KiB sent
  persistent keepalive: every 20 seconds
```

additional info from pi

edwardchenchen:
on my mac, I don't see the 192.168.1.0/24 in allowed ips
```
interface: utun0
  public key: removed
  private key: (hidden)
  listening port: 51232

peer: ivYjeZJeFN/ViyX4uHBWeJnCBK+WuLihYjI5NRNmT3w=
  endpoint: removed:56840
  allowed ips: 192.168.201.1/32, 192.168.201.2/32
  latest handshake: 50 seconds ago
[...]

PR #1270, opened 2022-06-29
add logic for calculating peers when relayed node is egress gateway

mattkasun:


0xdcarns (approved):

mattkasun:
fixes #1269

theguy951357:
@mattkasun @0xdcarns The egress can be pinged behind a relay, in front of a relay from a relayed server, and as a relaying server from both sides. the only instance where it cannot be pinged is if the .254 is relaying. other than that, everything else passes.

mattkasun:
@theguy951357 @0xdcarns --- as expected.  The problem with server as relay is covered by GRA-341

PR #1276, opened 2022-07-01
Feature v0.14.5 cert ha support

afeiszli:


afeiszli:
@mattkasun please review, this logic will add any key generated to the DB, and servers will only read their keys from the DB.

serverctl/tls.go
```diff
@@ -0,0 +1,105 @@
+package serverctl
```
mattkasun, line 1: why in package serverctl rather than adding to package tls

tls/tls.go
```diff
@@ -17,8 +17,23 @@ import (
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 )
 
-// CERTTIFICAT_VALIDITY duration of certificate validity in days
-const CERTIFICATE_VALIDITY = 365
+const (
```
mattkasun, line 20: some unused func in this package

serverctl/tls.go
afeiszli: The new functions Use the database package, and TLS package gets imported by netclient. This avoids adding the database package as a dependency to netclient

tls/tls.go
afeiszli: We saw the unused funcs but decided to leave them in in case you saw a need for them, but can take them out

PR #1298, opened 2022-07-06
add timeout for http requests

mattkasun:
timeout will prevent api calls from hanging
reason for server not responding to api call is still unknown.


0xdcarns (approved):

theguy951357:
@mattkasun @0xdcarns pass

PR #1302, opened 2022-07-06
reduce number of peer updates sent

0xdcarns:


mq/publishers.go
```diff
@@ -23,7 +23,7 @@ func PublishPeerUpdate(newNode *models.Node) error {
 	}
 	for _, node := range networkNodes {
 
-		if node.IsServer == "yes" {
+		if node.IsServer == "yes" || newNode.ID == node.ID { // skip servers and self
```
mattkasun, line 26: with this change, a newly joined node will not receive a peer update to up to 5 minutes.

mattkasun (changes requested):

PR #1316, opened 2022-07-06
Bugfix v0.14.5 wg ports

mattkasun:


0xdcarns (approved):

theguy951357:
on join, it says the "using port 0". on a netclient pull, it says "port is now 0". "wg show" shows normal listening ports and match up with other clients' "wg shows"

mattkasun:
Port 0 is correct for udp hole punch.

theguy951357:
ok cool. pass

PR #1389, opened 2022-07-13
Bash is a dependency for OpenWRT

HarvsG:
https://github.com/gravitl/netmaker/blob/ec90dab43736a7fff780442557e5a1880e62845e/scripts/netclient-install.sh#L246

### DIFF
--- a/scripts/netclient-install.sh
+++ b/scripts/netclient-install.sh
@@ -38,7 +38,7 @@ elif [ "${OS}" = "FreeBSD" ]; then
 	update_cmd='pkg update'
 	install_cmd='pkg install -y'
 elif [ -f /etc/openwrt_release ]; then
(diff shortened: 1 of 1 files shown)


theguy951357:
I am getting an error on here when I run the script
bash: -c: line 1: syntax error near unexpected token `do'
bash: -c: line 1: `do /sbin/netclient daemon  >> /tmp/netclient.logs 2>&1;           if [ 0 -gt 10240000 ];then tar zcf /tmp/netclient.logs.tar -C / tmp/netclient.logs  && > /tmp/netclient.logs;fi;done &'

afeiszli:
this is unrelated to the PR but likely needs to be solved. @HarvsG does this work on your setup?

0xdcarns (approved):

PR #1392, opened 2022-07-15
changed to always update server check-in times even when not leader

0xdcarns:
### To Test
- Run with more than 1 server
- Ensure server checkin times are always healthy/regularly updated


theguy951357 (approved):
checkins at healthy intervals from both servers. pass

mattkasun (approved):

abhishek9686 (approved):

PR #1398, opened 2022-07-19
Feature v0.14.6 addvalidation

cameronts:
Added validation to fix "Network.AddressRange6 has no validation"

Tested with go test ./...


mattkasun (changes requested):
test case InvalidAddress6 (lines 228-235 of network_test.go) should be uncommented.

PR #1400, opened 2022-07-19
fix_32_64_programfiles

si458:
this fixes windows 32 bit installs looking for wrong program files folder
but also moves the 64 build to the correct 64bit program files


si458:
> @si458 there are two problems we need to fix before merging this:
> 1. the installer sets up folders in x86
> 2. currently installed netclients will break, so we will need some form of upgrade logic

Depending on the installer you can get it to pick the correct program files folder (we use advanced installer and this is possible)

But as for the upgrade logic, I tried to learn go and failed after a few hurdles as it was way to complicated sorry 😞

PizzaProgram:
The solution is:
```
  fileName32 = os.Getenv("windir") + "\\System32\\wireguard.dll"
...
  path = os.Getenv("ProgramFiles") +  "\\Netclient\\"
```

PR #1401, opened 2022-07-19
Bugfix v0.14.5 etc hosts lockfile

mattkasun:
running multiple copies of netclient can corrupt /etc/hosts file.  Preventing multiple instances of netclient is not feasible as that would prevent running netclient join/pull/leave/install while netclient daemon is running.

added lockfile as a mutex for access to /etc/hosts  if lockfile exists, dns update is aborted.

to test:
- run multiple copies of netclient daemon and verify /etc/hosts is not corrupted 

  - in particular the part above
```
##################################################################
# Content under this line is handled by hostctl. DO NOT EDIT.
##################################################################


```
should not be updated

### DIFF
--- a/netclient/functions/mqhandlers.go
+++ b/netclient/functions/mqhandlers.go
@@ -247,9 +247,10 @@ func UpdatePeers(client mqtt.Client, msg mqtt.Message) {
 func setHostDNS(dns, iface string, windows bool) error {
 	etchosts := "/etc/hosts"
 	temp := os.TempDir()
-	lockfile := temp + "netclient-lock"
+	lockfile := temp + "/netclient-lock"
 	if windows {
 		etchosts = "c:\\windows\\system32\\drivers\\etc\\hosts"
+		lockfile = temp + "\\netclient-lock"
 	}
 	if _, err := os.Stat(lockfile); !errors.Is(err, os.ErrNotExist) {
 		return errors.New("/etc/hosts file is locked .... aborting")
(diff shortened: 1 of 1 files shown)


netclient/functions/mqhandlers.go
```diff
@@ -243,9 +246,20 @@ func UpdatePeers(client mqtt.Client, msg mqtt.Message) {
 
 func setHostDNS(dns, iface string, windows bool) error {
 	etchosts := "/etc/hosts"
+	temp := os.TempDir()
+	lockfile := temp + "netclient-lock"
```
0xdcarns, line 250: on your reference to temp + netclient-lock you have a "/" below on line 285, as well as `lockfile = temp + "\\netclient-lock"` under the windows condition on 289 (removeHostDNS func), whereas you do not on 250-253 in the setHostDNS func. Is that a concern that you're handling the lock file differently?

mattkasun: both functions should have the identical logic

0xdcarns (approved):

0xdcarns:
@theguy951357 rft

theguy951357 (approved):
got the message "[netclient] 2022-07-21 00:38:20 [mqhandlers.go-235] UpdatePeers(): error updating /etc/hosts /etc/hosts file is locked .... aborting" after running two instances on daemon pass

cameronts (approved):

PR #1426, opened 2022-07-27
Netmaker: Added config file handling for public IP service.

cameronts:


servercfg/serverconf.go
```diff
@@ -430,7 +430,13 @@ func GetPublicIP() (string, error) {
 	iplist := []string{"https://ip.server.gravitl.com", "https://ifconfig.me", "https://api.ipify.org", "https://ipinfo.io/ip"}
 	publicIpService := os.Getenv("PUBLIC_IP_SERVICE")
 	if publicIpService != "" {
-		logger.Log(3, "User provided public IP service is", publicIpService)
+		logger.Log(3, "User (environment variable) provided public IP service is", publicIpService)
+
+		// prepend the user-specified service so it's checked first
+		iplist = append([]string{publicIpService}, iplist...)
```
mattkasun, line 436: This results in two appends.  Line 442 also appends

config/environments/dev.yaml
```diff
@@ -11,3 +11,4 @@ server:
   disableremoteipcheck: "" # defaults to "false" or DISABLE_REMOTE_IP_CHECK (if set)
   version: "" # version of server
   rce: "" # defaults to "off"
+  publicIpService: "" # defaults to "" or PUBLIC_IP_SERVICE (if set)
```
mattkasun, line 14: Will need to write an update function, but that should be a separate PR

mattkasun (changes requested):

servercfg/serverconf.go
cameronts: Line 436 prepends for the environment variable case.  Line 442 is an else if and prepends for the config file case.

mattkasun: gotcha, should have opened the entire file instead of just looking at the github diff.... spacing was off and thought 442 was outside the if block

config/environments/dev.yaml
mattkasun: GRA-472

mattkasun (approved):

0xdcarns (approved):

PR #1436, opened 2022-08-02
Bugfix v0.14.7 wipelocal

abhishek9686:


netclient/functions/join.go
```diff
@@ -205,7 +205,7 @@ func JoinNetwork(cfg *config.ClientConfig, privateKey string) error {
 		logger.Log(0, "network:", cfg.Network, "failed to publish update for join", err.Error())
 	}
 
-	if cfg.Daemon == "install" || ncutils.IsFreeBSD() {
+	if cfg.Daemon == "on" || ncutils.IsFreeBSD() {
```
0xdcarns, line 208: this should remain "install", as packages handle installing the daemon generally

mattkasun, line 208: doc update required?

netclient/ncutils/netclientutils.go
```diff
@@ -592,3 +593,24 @@ func ModPort(node *models.Node) error {
 	}
 	return err
 }
+
+// CheckIfDaemonExists - checks if netclient daemon is up and running
+func CheckIfDaemonExists() bool {
+	daemonExists := false
+	pid, err := ReadPID()
```
0xdcarns, line 600: pid will not work for windows, for windows you will need to check for the presence of winsw.exe

0xdcarns (changes requested):

netclient/functions/join.go
abhishek9686: not required, changed it back to `install`

netclient/ncutils/netclientutils.go
abhishek9686: @0xdcarns found this one pkg (https://github.com/mitchellh/go-ps) to check for running process looks like it works for windows as well

abhishek9686: actually, this check wouldn't be needed anymore will remove it since the option is changed back to `install`

mattkasun (approved):

0xdcarns (approved):

theguy951357:
pass

PR #1449, opened 2022-08-08
Feature v0.14.7 translateiftablestonftables

cameronts:
Add support for nftables firewall on Linux.

### DIFF
--- a/logic/util.go
+++ b/logic/util.go
@@ -6,17 +6,17 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"math/big"
-	"math/rand"
-	"net"
-	"os"
-	"strings"
-	"time"
-
 	"github.com/gravitl/netmaker/database"
 	"github.com/gravitl/netmaker/logger"
 	"github.com/gravitl/netmaker/models"
 	"github.com/gravitl/netmaker/netclient/ncutils"
+	"math/big"
+	"math/rand"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+	"time"
 )
 
 // nfTablesPresent - returns true if nftables is present, false otherwise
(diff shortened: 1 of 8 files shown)


cameronts:
Sanity checks:

Run with iptables on all clients (the default)
See what PostUp and PostDown are in the UI for the ingress node 
	Before it is made an ingress node
In Netmaker UI
	On node’s command line (iptables)
	Make it an ingress node
	After it is made an ingress node
In Netmaker UI
	On node’s command line (iptables)
	Stop having it be an ingress node
	After it is no longer an ingress node
In Netmaker UI
	On node’s command line (iptables)
	Run with nft (apt install nftables, confirm installed by using nft list ruleset)
	On ingress node
	On another client node (in case commands are ending up on the non-ingress clients when running nft) (more of a “control” as it’s unlikely; just to be sure it’s not happening aka a negative test)
	See what PostUp and PostDown are in the UI for the ingress node 
		Before it is made an ingress node
			In Netmaker UI
			On node’s command line (iptables)
			On node’s command line (nft)
		Make it an ingress node
		After it is made an ingress node
			In Netmaker UI
			On node’s command line (iptables)
			On node’s command line (nft) 
Make it an egress node too
	See what the PostUp and PostDown commands are


If the nft rules are in place (shown by nft list ruleset)
	Create an ext client
	Connect the ext client (some other machine) to the ingress node
	Test pings and note what works to the various other nodes

0xdcarns (approved):

logic/util.go
```diff
@@ -6,17 +6,16 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/gravitl/netmaker/database"
```
mattkasun, line 9: order of imports

mattkasun (changes requested):
order of file imports

theguy951357:
pass

mattkasun (approved):

PR #1453, opened 2022-08-09
Fix: createExtClient()

DolevBaron:
The clientid supplied in the API call got discarded without ever being used, so createExtClient() could never be used with a custom client-id (until now).
I simply set the matching attribute before passing it to logic.CreateExtClient(), as otherwise it will always receive a blank ClientID & generate one on its own


mattkasun (changes requested):

afeiszli:
@DolevBaron the PR will not work as written because ClientID is not in the input params for creating an ext client. This will require more work in order to merge. Did you test?

k4s0:
I have crated a new PR with the working code for this request #1499

mattkasun:
superseded by #1499

PR #1460, opened 2022-08-17
move client public ip check to use server

afeiszli:
This commit adds a public ip checking service to the netmaker server. By doing so, clients can request their public IP from the netmaker server, instead of reaching out to an external service. 

The purpose of this is twofold:
1. It enables the 0.0.0.0/0 use case by having a reliable endpoint for all clients to reach to get their public IP which can be added to the routing table.
2. This should actually enable Netmaker to work in several countries where it does not currently, without needing the "additional allowedips" functionality


afeiszli:
To test (@theguy951357), spin up Netmaker with some clients. Check to see if they get the correct endpoints (should be fairly obvious if it is not working). Also, should test a client on the same server as Netmaker for safety, as well as clients that are in the same subnet (e.g. DigitalOcean region).

controllers/ipservice.go
```diff
@@ -0,0 +1,68 @@
...
+		return "", err
+	}
+	netIP = net.ParseIP(ip)
+	if netIP != nil {
+		if netIP.IsPrivate() || netIP.IsLoopback() {
+			return ip, fmt.Errorf("ip is a private address")
+		}
+		return ip, nil
+	}
+	return "", fmt.Errorf("No valid ip found")
```
mattkasun, line 67: as per the static check --- change to 'no valid ip found'

mattkasun (changes requested):
fix static check erros

controllers/ipservice.go
afeiszli: changed to "no ip found" on both

0xdcarns (approved):

mattkasun (changes requested):
script changes should be in separate PR.

mattkasun (approved):

mattkasun:
@theguy951357 RFT

theguy951357 (approved):
ips and ports look consistent. pass

PR #1466, opened 2022-08-19
add checks for firewall

mattkasun:
to test --- 

1. spin up fresh fedora 36/rocky linux droplet
2. install wireguard-tools
3. run ./netclient -v
4. verify error message 
`neither iptables nor nft is installed - node cannot be used as a gateway`
5. try to create ingress/egress on this node
6. UI error message

### DIFF
--- a/netclient/ncutils/netclientutils.go
+++ b/netclient/ncutils/netclientutils.go
@@ -113,15 +113,23 @@ func GetWireGuard() string {
 // IsNFTablesPresent - returns true if nftables is present, false otherwise.
 // Does not consider OS, up to the caller to determine if the OS supports nftables/whether this check is valid.
 func IsNFTablesPresent() bool {
-	nftFound := FileExists("/usr/sbin/nft")
-	logger.Log(3, "nftables found:", strconv.FormatBool(nftFound))
-	return nftFound
+	found := false
+	_, err := exec.LookPath("nft")
+	if err == nil {
+		found = true
+	}
+	return found
 }
 
 // IsIPTablesPresent - returns true if iptables is present, false otherwise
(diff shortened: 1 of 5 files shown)


netclient/ncutils/netclientutils.go
```diff
@@ -515,6 +515,22 @@ func CheckUID() {
...
+	_, err := exec.LookPath("iptables")
+	if err == nil {
+		found = true
+	}
+	_, err = exec.LookPath("nft")
+	if err == nil {
+		found = true
+	}
+	if !found {
+		log.Fatal("neither iptables nor nft is installed - please install one or the other and try again")
```
afeiszli, line 530: @mattkasun does this mean it will be mandatory to have either iptables or nft installed in order to run the netclient? Up until this point, I believe we've been able to run the netclient without iptables, and only needed it if doing ingress/egress.

mattkasun: currently yes, other option is to add a new field to nodes... canBeGateway or something similar .. would be set on startup and be limited to freebsd and linux with iptables or nft.   If flag is false - gateway creation in UI is disabled for that node

0xdcarns: @afeiszli we can leverage the firewall field on nodes and put an option as invalid and accomplish what matt is saying and then deny gateway creations on nodes that are invalid

theguy951357 (approved):
pass

netclient/ncutils/netclientutils.go
mattkasun: updated to match @0xdcarns comment

afeiszli (approved):

PR #1479, opened 2022-08-24
reverting mac path

afeiszli:


afeiszli:
The new mac path is causing a lot of issues running commands on mac. Have not found a solution so reverting ahead of the release, which will just have us use the standard path, which should be fine for now.

theguy951357 (approved):

mattkasun (approved):

PR #1481, opened 2022-08-25
freebsd saving post up/post down command to conf file

mattkasun:


netclient/wireguard/common.go
```diff
@@ -427,6 +440,7 @@ func UpdateWgPeers(file string, peers []wgtypes.PeerConfig) (*net.UDPAddr, error
 
 // UpdateWgInterface - updates the interface section of a wireguard config file
 func UpdateWgInterface(file, privateKey, nameserver string, node models.Node) error {
+	log.Println("updating conf file ", file, nameserver, node.Name)
```
afeiszli, line 443: should not use log.Println

netclient/wireguard/common.go
```diff
@@ -451,11 +465,25 @@ func UpdateWgInterface(file, privateKey, nameserver string, node models.Node) er
 	//if node.DNSOn == "yes" {
 	//	wireguard.Section(section_interface).Key("DNS").SetValue(nameserver)
 	//}
+	//need to split postup/postdown because ini lib adds a quotes which breaks freebsd
 	if node.PostUp != "" {
-		wireguard.Section(section_interface).Key("PostUp").SetValue(node.PostUp)
+		log.Println("updating PostUp")
```
afeiszli, line 470: remove log.Println

afeiszli (changes requested):
There are a couple of log.Printlns that should be removed or changed to logger.Log.

Also, in your comments for "spacing is important"; please clarify in comment exactly what format commands should take, so developers are aware in the future

afeiszli (approved):

PR #1497, opened 2022-08-30
add check for nil peer.Endpoint

mattkasun:
### Steps
* create an ingress gateway and ext client but don't connect ext client to network
* leave network or uninstall netclient on ingress node
### Expected Result
no panic


theguy951357 (approved):
pass

afeiszli (approved):

PR #1499, opened 2022-08-31
added new controller for creating custom ExtClient

k4s0:
Add the possibility to create custom clientid #1453


controllers/ext_client.go
```diff
@@ -248,9 +249,9 @@ func createExtClient(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
 	var params = mux.Vars(r)
-
 	networkName := params["network"]
 	nodeid := params["nodeid"]
+	clientid := params["clientid"]
```
mattkasun, line 254: client id should be retrieved from struct passed in request.Body not from url parameter

mattkasun (changes requested):

controllers/ext_client.go
k4s0: OK but i need to use the ExtClient struct or create a new struct for this type of request?

mattkasun: you could do either ... use the ExtClient struct but only send the ClientID param in the body of the request or use a new struct.
I would create an anonymous struct consisting of just one element and decode the body into that struct and then copy to extclient struct.  
eg.

request := struct {
     id string
}
if err := json.NewDecoder(r.body).Decode(&request); err != nil {
   //log error
} else {
   extclient.ClientID = request.id
}

k4s0: I have made the changes

controllers/ext_client.go
```diff
@@ -261,6 +261,16 @@ func createExtClient(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var extclient models.ExtClient
+	var CustomExtClient models.CustomExtClient
+	
+	err := json.NewDecoder(r.Body).Decode(&CustomExtClient);
+
+	if err != nil {
+		logger.Log(1, "error creating CustomExtClient"+err.Error())
```
0xdcarns, line 269: log message suggests an error occurred, but it's more of an absence of a clientid, would either log that no clientid was provided and one will be created, or just check if err == nil { extclient.ClientID = CustomExtClient.ClientID } (no log)

0xdcarns (changes requested):

controllers/ext_client.go
k4s0: @0xdcarns ok you're right i have made the changes

mattkasun (approved):

afeiszli:
@k4s0 is 'lorenzo' also you in the commit? Both are showing up as commiters so CLA Assistant is looking for both.

Once that is sorted we will merge.

k4s0:
> @k4s0 is 'lorenzo' also you in the commit? Both are showing up as commiters so CLA Assistant is looking for both.
> 
> Once that is sorted we will merge.

Yes both mine because i used two different git config, in two different machine, but just only one is a github users, so how can i resolve this?

afeiszli (approved):

0xdcarns (approved):

PR #1506, opened 2022-09-01
fixing ip check for server

afeiszli:


afeiszli:
egressIPs[i].IP.String() returns "0.0.0.0" so never catches
egressIPs[i].String() returns "0.0.0.0/0" and is correct

mattkasun (approved):

0xdcarns (approved):

PR #1536, opened 2022-09-08
send connected status on checkin

mattkasun:


0xdcarns:
there is probably a better way to handle this.

netclient/functions/connection.go
```diff
@@ -23,6 +24,7 @@ func Connect(network string) error {
 	if err = wireguard.ApplyConf(&cfg.Node, cfg.Node.Interface, filePath); err != nil {
 		return err
 	}
+	daemon.Restart()
```
0xdcarns, line 27: we would probably want to call this last, after modding the config.

theguy951357:
journalctl is showing healthy checkins and peer updates, but the UI shows nodes going into warning

0xdcarns (approved):

PR #1539, opened 2022-09-09
use RunCmd for postup/postdown

mattkasun:
To test run on distro without wg-quick, eg. openwrt

after joining, create gateway(ingress or egress) and view logs for absence of panic

### DIFF
--- a/netclient/ncutils/netclientutils.go
+++ b/netclient/ncutils/netclientutils.go
@@ -437,6 +437,10 @@ func Copy(src, dst string) error {
 func RunCmds(commands []string, printerr bool) error {
 	var err error
 	for _, command := range commands {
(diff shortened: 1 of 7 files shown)


netclient/ncutils/netclientutils.go
```diff
@@ -437,6 +437,10 @@ func Copy(src, dst string) error {
 func RunCmds(commands []string, printerr bool) error {
 	var err error
 	for _, command := range commands {
+		//prevent panic
+		if command == " " {
```
0xdcarns, line 441: could probably do a trim and check if length == 0

abhishek9686 (approved):

0xdcarns (approved):

PR #1559, opened 2022-09-14
Feature v0.16.0 ee update

afeiszli:


auth/oidc.go
```diff
@@ -67,14 +62,16 @@ func handleOIDCLogin(w http.ResponseWriter, r *http.Request) {
 		http.Redirect(w, r, servercfg.GetFrontendURL()+"/login?oauth=callback-error", http.StatusTemporaryRedirect)
 		return
 	}
-
+	logger.Log(3, "using state string:", oauth_state_string)
```
0xdcarns, line 65: Remove this log.

auth/google.go
```diff
@@ -94,10 +91,10 @@ func handleGoogleCallback(w http.ResponseWriter, r *http.Request) {
 	http.Redirect(w, r, servercfg.GetFrontendURL()+"/login?login="+jwt+"&user="+content.Email, http.StatusPermanentRedirect)
 }
```
abhishek9686, line 92: can we do a `fmt.Sprintf` here for the redirect url, it will be much cleaner

auth/nodecallback.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	// all responses should be in html format from here on out
+	w.Header().Add("content-type", "text/html; charset=utf-8")
+
+	// retrieve machinekey from state cache
+	reqKeyIf, machineKeyFoundErr := netcache.Get(state)
+	if machineKeyFoundErr != nil {
+		logger.Log(0, "requested machine state key expired before authorisation completed -", err.Error())
+		reqKeyIf = &netcache.CValue{
```
abhishek9686, line 62: need to use `machineKeyFoundErr` in the log statement

auth/nodesession.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/gravitl/netmaker/models"
+	"github.com/gravitl/netmaker/models/promodels"
+	"github.com/gravitl/netmaker/servercfg"
+)
+
+// SessionHandler - called by the HTTP router when user
+// is calling netclient with --login-server parameter in order to authenticate
+// via SSO mechanism by OAuth2 protocol flow.
+// This triggers a session start and it is managed by the flow implmented here and callback
+// When this method finishes - the auth flow has finished either OK or by timeout or any other error occured
```
abhishek9686, line 23: typo `implmented` ---> `implemented`

auth/nodesession.go
```diff
@@ -0,0 +1,155 @@
...
+		logger.Log(0, "failed to remove node SSO cache entry", err.Error())
+	}
+	// Cleanly close the connection by sending a close message and then
+	// waiting (with timeout) for the server to close the connection.
+	err = conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
+	if err != nil {
+		logger.Log(0, "write close:", err.Error())
+		return
+	}
+	time.After(time.Second)
```
abhishek9686, line 152: why is this `1s` wait required?

auth/nodesession.go
```diff
@@ -0,0 +1,155 @@
...
+	close(answer)
+	close(timeout)
+}
```
abhishek9686, line 155: you can close the channels in the `defer func()` otherwise any error above before this line will return without closing the channels

controllers/limits.go
```diff
@@ -0,0 +1,58 @@
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		var errorResponse = models.ErrorResponse{
+			Code: http.StatusUnauthorized, Message: "free tier limits exceeded on networks",
+		}
+
+		if logic.Free_Tier && logic.Is_EE { // check that free tier limits not exceeded
+			if limit_choice == networks_l {
+				currentNetworks, err := logic.GetNetworks()
+				if (err != nil && !database.IsEmptyRecord(err)) || len(currentNetworks) >= logic.Networks_Limit {
+					logic.ReturnErrorResponse(w, r, errorResponse)
```
abhishek9686, line 29: shouldn't the limit check should be `>` rather than `>=`

controllers/network.go
```diff
@@ -199,21 +193,39 @@ func updateNetwork(w http.ResponseWriter, r *http.Request) {
...
+	rangeupdate4, rangeupdate6, localrangeupdate, holepunchupdate, groupsDelta, userDelta, err := logic.UpdateNetwork(&network, &newNetwork)
 	if err != nil {
 		logger.Log(0, r.Header.Get("user"), "failed to update network: ",
 			err.Error())
-		returnErrorResponse(w, r, formatError(err, "badrequest"))
+		logic.ReturnErrorResponse(w, r, logic.FormatError(err, "badrequest"))
 		return
 	}
 
+	if len(groupsDelta) > 0 {
```
abhishek9686, line 204: tbh you don't need this `len` check here

mq/publishers.go
```diff
@@ -176,3 +181,67 @@ func ServerStartNotify() error {
...
+					}
+				}
+			}
+		}
+	}
+}
+
+func pushMetricsToExporter(metrics models.Metrics) error {
+	logger.Log(2, "----> Pushing metrics to exporter")
+	SetupMQTT()
```
abhishek9686, line 232: I guess we can remove `SetupMQTT()` here since mqclient is a global variable now

netclient/functions/join.go
```diff
@@ -8,21 +8,174 @@ import (
...
+	for {
+		select {
+		case <-done:
+			logger.Log(1, "finished")
+			return nil
+		case <-interrupt:
+			log.Println("interrupt")
+			// Cleanly close the connection by sending a close message and then
+			// waiting (with timeout) for the server to close the connection.
+			err := conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
```
abhishek9686, line 165: you can add an additional case for WebSocket close message type from the server and then close the connection on the client side

netclient/functions/mqpublish.go
```diff
@@ -146,6 +160,78 @@ func Hello(nodeCfg *config.ClientConfig) {
...
 
+// publishMetrics - publishes the metrics of a given nodecfg
+func publishMetrics(nodeCfg *config.ClientConfig) {
+	token, err := Authenticate(nodeCfg)
+	if err != nil {
+		logger.Log(1, "failed to authenticate when publishing metrics", err.Error())
+		return
+	}
+	url := "https://" + nodeCfg.Server.API + "/api/nodes/" + nodeCfg.Network + "/" + nodeCfg.Node.ID
+	response, err := API("", http.MethodGet, url, token)
```
abhishek9686, line 171: use `fmt.Sprintf` for url

controllers/limits.go
0xdcarns: no because that would imply they were able to make more than the limit previously

controllers/network.go
0xdcarns: those delta slices have the potential to be nil

controllers/limits.go
abhishek9686: oh got it my bad that is a middleware func which checks before creating resources, makes sense

abhishek9686 (approved):

theguy951357:
all good

0xdcarns (approved):

PR #1596, opened 2022-09-22
ee check

mattkasun:
To test:  verify no "collecting metrics" in netclient logs when connected to non ee server

### DIFF
--- a/models/structs.go
+++ b/models/structs.go
@@ -218,6 +218,7 @@ type ServerConfig struct {
 	Version     string `yaml:"version"`
 	MQPort      string `yaml:"mqport"`
 	Server      string `yaml:"server"`
+	Is_EE       bool   `yaml:"isee"`
 }
 
 // User.NameInCharset - returns if name is in charset below or not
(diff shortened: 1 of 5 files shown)


afeiszli (approved):

0xdcarns:
in servercfg/serverconf.go:92, the GetServerInfo func does not return/set if it is indeed in ee mode or not. In order to do this, you'll need the same logic from controllers/server.go:92-95

models/structs.go
```diff
@@ -218,6 +218,7 @@ type ServerConfig struct {
 	Version     string `yaml:"version"`
 	MQPort      string `yaml:"mqport"`
 	Server      string `yaml:"server"`
+	Is_EE       bool   `yaml:"isee"`
```
0xdcarns, line 221: see comment in PR

0xdcarns (changes requested):

models/structs.go
mattkasun: moved IS_EE var to servercfg.

0xdcarns (approved):

PR #1607, opened 2022-09-26
Bugfix v0.16.1 ingress egress

mattkasun:


0xdcarns (approved):

agorgl:
This bug prevents netmaker server to act both as an ingress and egress currently, do we expect a hotfix release soon?

mattkasun:
> This bug prevents netmaker server to act both as an ingress and egress currently, do we expect a hotfix release soon?

no, bug only happens if you create ingress and then egress.  If you created egress and then ingress, the bug does not happen.

theguy951357:
I cannot ping the ext client on this one.

afeiszli (approved):

PR #1609, opened 2022-09-27
add egress range to allowed ips if relay of relayed node is egress ga…

mattkasun:
…teway

### DIFF
--- a/logic/peers.go
+++ b/logic/peers.go
@@ -466,6 +466,17 @@ func GetPeerUpdateForRelayedNode(node *models.Node, udppeers map[string]string) 
 			}
 		}
 	}
+	//add egress range if relay is egress
+	if relay.IsEgressGateway == "yes" {
+		var ip *net.IPNet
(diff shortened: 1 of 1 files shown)


afeiszli (approved):

theguy951357 (approved):
pass

0xdcarns (approved):

logic/peers.go
```diff
@@ -466,6 +466,17 @@ func GetPeerUpdateForRelayedNode(node *models.Node, udppeers map[string]string)
...
+	//add egress range if relay is egress
+	if relay.IsEgressGateway == "yes" {
+		var ip *net.IPNet
+		for _, cidr := range relay.EgressGatewayRanges {
+			_, ip, err = net.ParseCIDR(cidr)
+			if err != nil {
+				continue
+			}
+		}
+		allowedips = append(allowedips, *ip)
```
0xdcarns, line 478: having the append after the loop will result in only the last CIDR being appended correct?

Issue #1611, opened 2022-09-28
[Bug]: Key: 'Network.NetID' Error:Field validation for 'NetID' failed on the 'netid_valid' tag


alcroito:
### What happened?

A bug happened!

Hi. recently I upgraded the netmaker server from v0.11 to 0.16, and upon start up of the netmaker docker container i always get an error as shown in the log snippet.
Connections ends up working, but trying to edit and save the network settings causes a panic, as shown in the log snippet.

Perhaps I can modify the sqlite database directly to fix this?

### Version

v0.16.0

### What OS are you using?

Linux

### Relevant log output

```shell
netmaker       | [netmaker] 2022-09-28 01:23:07 connecting to sqlite
netmaker       | [netmaker] 2022-09-28 01:23:07 database successfully connected
netmaker       | [netmaker] 2022-09-28 01:23:07 no OAuth provider found or not configured, continuing without OAuth
netmaker       | Key: 'Network.NetID' Error:Field validation for 'NetID' failed on the 'netid_valid' tag
netmaker       | Key: 'Network.DefaultACL' Error:Field validation for 'DefaultACL' failed on the 'checkyesorno' tag
netmaker       | [netmaker] 2022-09-28 01:23:07 could not set defaults on network XTSYakwp
netmaker       | Key: 'Network.DefaultACL' Error:Field validation for 'DefaultACL' failed on the 'checkyesorno' tag
netmaker       | [netmaker] 2022-09-28 01:23:07 could not set defaults on network my-network

...

netmaker       | 2022/09/28 01:26:45 http: panic serving 172.21.0.1:55530: runtime error: invalid memory address or nil pointer dereference
netmaker       | goroutine 278 [running]:
netmaker       | net/http.(*conn).serve.func1()
netmaker       | 	/usr/local/go/src/net/http/server.go:1825 +0xbf
netmaker       | panic({0xaf1c80, 0x100f030})
netmaker       | 	/usr/local/go/src/runtime/panic.go:844 +0x258
netmaker       | github.com/gravitl/netmaker/logic.UpdateNetwork(0xc0004833f8, 0xc000346a00)
netmaker       | 	/app/logic/networks.go:559 +0x1c8
netmaker       | github.com/gravitl/netmaker/controllers.updateNetwork({0xc948d0, 0xc000450460}, 0xc00045af00)
netmaker       | 	/app/controllers/network.go:196 +0x36c
[...]

PR #1613, opened 2022-09-28
check netid on start

mattkasun:
- compare netID on during initialization and update netID/DefautlInterface if contains uppercase or a . 
- setDefaults for networks

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,6 +12,7 @@ import (
 	"os/signal"
 	"runtime/debug"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


mattkasun:
fixes #1611

theguy951357 (approved):
good

main.go
```diff
@@ -74,6 +75,38 @@ func initialize() { // Client Mode Prereq Check
 	if err = database.InitializeDatabase(); err != nil {
 		logger.FatalLog("Error connecting to database")
 	}
+	//update network defaults
+	networks, _ := logic.GetNetworks()
```
0xdcarns, line 79: this whole block should probably be in another function

0xdcarns: further, you could add this logic to setNetworkDefaults in `serverctl.go:124`

serverctl/serverctl.go
```diff
@@ -132,9 +132,32 @@ func setNetworkDefaults() error {
...
+			newNet.SetDefaults()
+			if err := logic.SaveNetwork(&newNet); err != nil {
+				logger.Log(0, "error saving networks during initial update:", err.Error())
+			}
+			if err := logic.DeleteNetwork(net.NetID); err != nil {
+				logger.Log(0, "error deleting old network:", err.Error())
+			}
+		} else {
+			net.SetDefaults()
+			_, _, _, _, _, _, err = logic.UpdateNetwork(&net, &net)
```
0xdcarns, line 157: should this not just be logic.SaveNetwork? since you're updating with the same network struct?

mattkasun: Update network calls validation

0xdcarns: ah okay, probably need to add that to save in the future then

0xdcarns (approved):

afeiszli (approved):

PR #1644, opened 2022-10-09
fix workflow

afeiszli:


.github/workflows/publish-netclient-docker.yml
```diff
@@ -71,28 +71,29 @@ jobs:
...
-          tags: ${{ env.TAG }}
-          build-args: version=${{ env.TAG }}  
-      -
-        name: Test armv7l
-        run: |
-            docker run --rm ${{ env.TAG }}&
-            sleep 10
-            kill %1
+      # -
+      #   name: Build armv7l and export to Docker
```
mattkasun, line 75: Remove instead of commenting out

mattkasun (changes requested):

.github/workflows/publish-netclient-docker.yml
afeiszli: removed

mattkasun (approved):

Issue #1646, opened 2022-10-10
Netclient defaults to nfttables for FW, causes iptables machine to break


LucaOonk:
What happened?

When adding a new Node it defaults to use the `nfttables` in `firewallinuse`, even though the active firewall is `iptables`. When this node is made an ingress-node this machine becomes unresponsive,  and can not be reached. Requiring the following to make the machine work again:
- Reboot the machine
- Netclient remove the network as soon as the machine responds again, hoping it wont break again.
- Reboot the machine again

The workaround I found was make an api call to change the `firewallinuse` and then make the node an ingress:

```
curl 'https://api.<DOMAIN>/api/nodes/<NETWORK>/<NODEID>' \
  -X 'PUT' \
  -H 'accept: application/json, text/plain, */*' \
  -H 'accept-language: nl-NL,nl;q=0.9,en-US;q=0.8,en;q=0.7' \
  -H 'authorization: Bearer <TOKEN>' \
  -H 'content-type: application/json' \
  -H 'sec-fetch-mode: cors' \
  -H 'user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/105.0.0.0 Safari/537.36' \
  --data-raw '{"firewallinuse":"iptables"}' \
  --compressed
```

My guess is that this is not intended and could be prevented. For example, check the current firewall when the node is added.
Maybe this kind of advanced info could be shown in a special tab for each node? This would greatly help troubleshoot stuff like this!

Version
v0.16.0 server
v0.16.0 netclient

What OS are you using?
Linux

Relevant log output
No response

Contributing guidelines
 Yes, I did.

mattkasun:
netclient defaults to iptables unless nftables is present on the host, assumption being that latest firewall tool installed is to be used.  This assumption may need to be revisited.

LucaOonk:
If I understand correctly, when nfttables is deleted before enrolling the node it will use iptables, correct?

I had some idea's how to "fix" this:
- Maybe this could be "fixed" by adding a promt when a new node is added? Then the admin, who has this info, cloud correctly configure it without breaking the server+ssh (and other services).
- Or being able to set/change this as a network/node setting using the Netmaker UI

LucaOonk:
Hi @mattkasun,

[...]

PR #1696, opened 2022-10-28
Added user FW selection promt Fix for issue #1646

LucaOonk:
# The problem:
In the current versions of Netclient it is opssible to brick the device when configuring conflicting firewalls to use. For example:
The main FW is IPtables, but NFTables is also installed. Due to the fact that netclient prefers NFTables, there is a situation where the machine would become unresponsive. 

# The solution:
I have added a user promt to fix an issue as described in: [https://github.com/gravitl/netmaker/issues/1646](https://github.com/gravitl/netmaker/issues/1646)

When enrolling a client, using the join command and both NFTables and IPTables are present on the machine, the user will be promted to set their preffered FW to use. 

This promt has 2 options:
(1) NFTables
(2) IPTables

If something else is entered it will default to use FW NONE.

I hope this will save somebody's day.

### DIFF
--- a/netclient/functions/join.go
+++ b/netclient/functions/join.go
@@ -277,13 +277,39 @@ func JoinNetwork(cfg *config.ClientConfig, privateKey string) error {
 	}
 
 	if cfg.Node.FirewallInUse == "" {
-		if ncutils.IsNFTablesPresent() {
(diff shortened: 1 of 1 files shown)


theguy951357 (approved):
pass

mattkasun (approved):

PR #1769, opened 2022-11-28
add all local interfaces to node

mattkasun:


models/node.go
```diff
@@ -43,6 +43,12 @@ var seededRand *rand.Rand = rand.New(
 type NodeCheckin struct {
 	Version   string
 	Connected string
+	Ifaces    []Iface
+}
+
+type Iface struct {
```
0xdcarns, line 49: missing comment for public type

0xdcarns (changes requested):

netclient/functions/join.go
```diff
@@ -238,6 +238,14 @@ func JoinNetwork(cfg *config.ClientConfig, privateKey string) error {
 			logger.Log(1, "network:", cfg.Network, "error retrieving private address: ", err.Error())
 		}
 	}
+	if len(cfg.Node.Interfaces) == 0 {
+		ip, err := getInterfaces()
+		if err != nil {
+			logger.Log(0, "failed to retrive local interfaces", err.Error())
```
0xdcarns, line 244: if there is an error, should we set interfaces to an empty array, in-case it's nil?

0xdcarns: same to below file

mattkasun: the only way this interfaces can get set to nil is if getInterfaces() returns nil,nil , ie no errors retrieving interfaces but no actual interfaces.  If someone is running netclient with no interfaces a panic isn't out of the question.

if the case of an error from getInterfaces(), continuing to use the previously generated array is less misleading than setting interfaces to an empty array.

0xdcarns (approved):

alphadose (approved):

abhishek9686 (approved):

PR #1797, opened 2022-12-07
Fixes+Chores: avoid de-referencing nil ptrs + lint

yunginnanet:
1) Avoid referencing conditions we know are false/true

 2) Avoid using name of imported package as variable

 3) Avoid broken (see list item 1) if else statement in `ipservice.go` by refactoring to switch statement

 4) When assigning a pointer value to a variable along with an error, check that error before referencing that pointer. **Thus avoiding de-referencing a nil and causing a panic**. (_This list item is by far the most important_)

 5) Standard gofmt package sorting + linting; This includes fixing comment starts for go doc

 6) Explicit non-handling of unhandled errors where appropriate (assigning errs to _ to reduce linter screaming)

 7) Export ErrExpired in `netcache` package so that we can properly reference it using `errors.Is` instead of using `strings.Contains` against an `error.Error()` value
 
 ---
 
 The project still doesn't quite pass `go vet ./...` unfortunately due to excessive usage of unkeyed fields when substantiating instances of defined types, but I already felt like this was a sizable commit to review without tacking that on.

### DIFF
--- a/netclient/ncutils/netclientutils.go
+++ b/netclient/ncutils/netclientutils.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/c-robinson/iplib"
+
 	"github.com/gravitl/netmaker/logger"
 	"github.com/gravitl/netmaker/models"
 	"github.com/gravitl/netmaker/netclient/global_settings"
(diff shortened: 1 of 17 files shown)


netclient/ncutils/netclientutils.go
```diff
@@ -167,13 +168,15 @@ func GetPublicIP(api string) (string, error) {
...
+		var resp *http.Response
+		resp, err = client.Get(ipserver)
 		if err != nil {
 			continue
 		}
 		defer resp.Body.Close()
 		if resp.StatusCode == http.StatusOK {
-			bodyBytes, err := io.ReadAll(resp.Body)
+			var bodyBytes []byte
+			bodyBytes, err = io.ReadAll(resp.Body)
```
yunginnanet, line 179: note that while this fixes the previous problem where one was handling an error outside of the for loop that wasn't in it's context - upon handling the error type outside of the for loop we're only going to know about the last error - not any of the others  
  
\#### tldr;   
it's better now but this for loop still has it's problems

0xdcarns (approved):

alphadose (approved):

Issue #1764, opened 2022-11-25
[Bug]: Ext. Clients & Nodes naming allows characters which break functionality


pipelinedave:
### Contact Details

kingovi@gmail.com

### What happened?

if you assign a name that includes for example a slash ("/") and then try to edit that client, the edit form will not show up.
web console also gives a 401 error.
i suspect this is because it seems to break the path in the url.

have not tested other scenarios, but i suspect it would misbehave in a similar way.

### Version

v0.16.3

### What OS are you using?

Linux

### Contributing guidelines

- [X] Yes, I did.

PR #1804, opened 2022-12-08
Bugfix allow only alphanumeric and dashes for ext client and node names

alphadose:
Fixes https://github.com/gravitl/netmaker/issues/1764

### DIFF
--- a/controllers/node.go
+++ b/controllers/node.go
@@ -538,6 +538,10 @@ func createNode(w http.ResponseWriter, r *http.Request) {
 		logic.ReturnErrorResponse(w, r, logic.FormatError(err, "badrequest"))
 		return
 	}
+	if !validName(node.Name) {
+		logic.ReturnErrorResponse(w, r, logic.FormatError(errInvalidNodeName, "badrequest"))
+		return
+	}
 
 	node.Network = networkName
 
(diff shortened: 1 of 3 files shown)


controllers/node.go
```diff
@@ -538,6 +538,10 @@ func createNode(w http.ResponseWriter, r *http.Request) {
 		logic.ReturnErrorResponse(w, r, logic.FormatError(err, "badrequest"))
 		return
 	}
+	if !validName(node.Name) {
```
mattkasun, line 541: node.Name is already validated in logic.CreateNode and logic.Update node...  This validation is redundant

mattkasun (changes requested):

Aceix (approved):

controllers/node.go
Aceix: @mattkasun  this Pr is as a result of this issue https://github.com/gravitl/netmaker/issues/1764

Aceix: the resolution we agreed on https://toil.kitemaker.co/cQARgD-Gravitl/Nk7Fgo-Community/items/753

0xdcarns (approved):

controllers/node.go
mattkasun: try and change a node name to something other than char or - on existing server .... It will fail ... the nodes checks are not needed

alphadose: @mattkasun removed the redundancy

mattkasun (approved):

PR #1808, opened 2022-12-09
fixed ee user pwd change err, fixed ee license free tier limit err

0xdcarns:


alphadose (approved):

ee/initialize.go
```diff
@@ -48,6 +48,9 @@ func setControllerLimits() {
 	logic.Clients_Limit = Limits.Clients
 	logic.Free_Tier = Limits.FreeTier
 	servercfg.Is_EE = true
+	if logic.Free_Tier {
+		logic.Networks_Limit = 3
```
mattkasun, line 52: Should this be a constant

mattkasun (changes requested):

ee/initialize.go
0xdcarns: It's only used once, generally constants are made if values are necessary if used more than once.

mattkasun (approved):

PR #1822, opened 2022-12-14
fix(GRA-752f): add address string to be sent to frontend

Aceix:
Add a field for the IP address to be sent to frontend as string. ref: https://pkg.go.dev/net#IPNet.String


mattkasun (commented):
Not sure that this is the best way to do this.  The conversion of an address to a string for the UI should be done on server not on client.

afeiszli (approved):

alphadose (approved):

PR #1833, opened 2022-12-19
new host model

mattkasun:


models/host.go
```diff
@@ -0,0 +1,33 @@
+package models
+
+import (
+	"net"
+
+	"github.com/google/uuid"
+	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
+)
+
+type Host struct {
```
0xdcarns, line 10: missing comment for public type

0xdcarns (changes requested):

0xdcarns (approved):

PR #1837, opened 2022-12-20
Sort peers wrt public key before publishing

alphadose:


0xdcarns:
See https://github.com/gravitl/netmaker/pull/1838, this could work, but to be more specific, it was better to sort right before returned peers

PR #1841, opened 2022-12-20
Gra 824 refactor join

mattkasun:
also removes server node ----

proxies are likely broken and UI will probably require updates to display node data

unit tests are broken


logic/metrics/metrics.go
```diff
@@ -102,7 +102,7 @@ func GetExchangedBytesForNode(node *models.Node, metrics *models.Metrics) error
 		return err
 	}
 	defer wgclient.Close()
-	device, err := wgclient.Device(node.Interface)
+	device, err := wgclient.Device("netmaker")
```
0xdcarns, line 105: netmaker should be a constant

0xdcarns (changes requested):

logic/metrics/metrics.go
mattkasun: added new constant; applied to other instances as well

0xdcarns (approved):

PR #1848, opened 2022-12-21
created conversion functions for hosts

0xdcarns:


models/api_host.go
```diff
@@ -1,27 +1,114 @@
...
+	a.PublicKey = h.PublicKey.String()
+	a.Verbosity = h.Verbosity
+	a.Version = h.Version
+
+	return &a
+}
+
+// APIHost.ConvertAPIHostToNMHost - convert's a given apihost struct to
+// a Host struct
+func (a *ApiHost) ConvertAPIHostToNMHost(currentHost *Host) *Host {
```
Aceix, line 69: why does this need a parameter?

0xdcarns: there are fields which are not editable, but we don't send to the UI, so we need to ensure they stay on the server

Aceix (approved):

PR #1857, opened 2022-12-22
updated controllers to use conversion routines

0xdcarns:


controllers/node.go
```diff
@@ -871,13 +877,109 @@ func deleteIngressGateway(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+	if relayedUpdate {
+		updateRelay(&currentNode, &newNode)
+	}
+	if servercfg.IsDNSMode() {
+		logic.SetDNS()
+	}
+
+	apiNode := newNode.ConvertToAPINode()
```
mattkasun, line 975: why return api node?  if called from node, the node would expect a clientNode back

mattkasun (changes requested):

mattkasun (approved):

PR #1865, opened 2022-12-23
Make FRONTEND_URL param required only for UI login

alphadose:
The `FRONTEND_URL` is still required for successful login via redirect on oauth callbacks https://github.com/gravitl/netmaker/blob/master/auth/github.go#L89


Aceix:
Will check and add pass the frontend redirect URL

0xdcarns (approved):

mattkasun (approved):

PR #1881, opened 2022-12-28
made mq functions a little easier to read and re-use

0xdcarns:


abhishek9686 (approved):

abhishek9686:
in the delete host API handler, you need to delete the MQ client for the host

PR #1884, opened 2022-12-29
send node update only if left from UI, modify mq client if host already exists

abhishek9686:


controllers/node.go
```diff
@@ -592,12 +594,25 @@ func createNode(w http.ResponseWriter, r *http.Request) {
...
 	}
+
 	host, err := logic.GetHost(data.Host.ID.String())
 	if err != nil {
 		logger.Log(0, r.Header.Get("user"), "failed to find host:", err.Error())
 		logic.ReturnErrorResponse(w, r, logic.FormatError(err, "internal"))
 		return
 	}
+	if modifyMqClient {
+		networks := logic.GetHostNetworks(data.Host.ID.String())
```
0xdcarns, line 605: why not just put this statement on line 578

0xdcarns (changes requested):

0xdcarns (approved):

PR #1904, opened 2023-01-05
Add space after double slash in comment

ksovi:
The build breaks here if the comment is not properly formatted


Aceix:
It is a go directive. There shouldn't be a space. Rather, you would need to create the file in the right directory

mattkasun (approved):

PR #1906, opened 2023-01-06
disallowed host removal if nodes not dissociated

0xdcarns:


Aceix (approved):

controllers/node.go
```diff
@@ -1064,24 +1064,34 @@ func deleteNode(w http.ResponseWriter, r *http.Request) {
...
+		var foundNode *models.Node
 		for _, nodetocheck := range nodes {
 			if nodetocheck.HostID == node.HostID {
-				foundNode = nodetocheck
+				foundNode = &nodetocheck
 				break
 			}
 		}
+		if foundNode == nil { // check if node is in deleted nodes if not found
+			delNode, err := logic.GetDeletedNodeByID(node.ID.String())
```
abhishek9686, line 1075: Node deletion already happens at line 1053, i have fixed this already in my current PR #1905

0xdcarns: this is the case in which a node receives a DELETE action and responds itself rather the API

mattkasun (approved):

PR #1939, opened 2023-01-16
ignored nodes with PendingDelete set

0xdcarns:
To test:
- Join at least 2 nodes
- Turn off daemon on one of the hosts
- Delete that host's node via API (UI or NMCTL).
- The other should no longer have that node in it's peers


mattkasun (approved):

Aceix (approved):

theguy951357 (approved):
joined 2 nodes to dev
shut down daemon on ubuntuQA
removed node through nmctl node delete dev <ubuntuQA id>
wg on fedoraQA shows peer is gone
pass

PR #1940, opened 2023-01-16
mq handler for host/delete

mattkasun:


abhishek9686 (commented):
should be implemented after host updates are done GRA-985

theguy951357 (approved):
made netcliet on ubuntu, fedora, and freebsd
ran netclient uninstall
/etc/netclient is empty
find netclient in sbin shows the binary is gone
on freebsd /var/log/netclient.log is gone
netmaker interface is gone
server shows nodes and hosts gone
pass

mq/handlers.go
```diff
@@ -118,6 +118,39 @@ func UpdateNode(client mqtt.Client, msg mqtt.Message) {
...
+// DeleteHost message Handler -- handles host deletion messages
+func DeleteHost(client mqtt.Client, msg mqtt.Message) {
+	id, err := getID(msg.Topic())
+	if err != nil {
+		logger.Log(1, "error getting host.ID sent on ", msg.Topic(), err.Error())
+		return
+	}
+	host, err := logic.GetHost(id)
+	if err != nil {
+		logger.Log(1, "error getting node ", id, err.Error())
```
Aceix, line 130: should be `error getting host`

mq/handlers.go
```diff
@@ -118,6 +118,39 @@ func UpdateNode(client mqtt.Client, msg mqtt.Message) {
...
+			continue
+		}
+		if err := logic.DeleteNode(&node, true); err != nil {
+			logger.Log(0, "failed to delete node", node.ID.String(), err.Error())
+			continue
+		}
+		logger.Log(3, "deleted node", node.ID.String(), "to facilitate deletion of host", host.ID.String())
+	}
+	if err := logic.RemoveHost(host); err != nil {
+		logger.Log(0, "error removing host", id, err.Error())
```
Aceix, line 146: would you want to end prematurely and not publish to mq since the host didnt get deleted?

Aceix (changes requested):

0xdcarns:
Has been made with completion of ticket GRA-985fg

PR #1949, opened 2023-01-18
Gra 985/host listen port

abhishek9686:
@theguy951357 refer PR https://github.com/gravitl/netclient/pull/139 for instructions

### DIFF
--- a/models/api_host.go
+++ b/models/api_host.go
@@ -48,7 +48,6 @@ func (h *Host) ConvertNMHostToAPI() *ApiHost {
 	}
 	a.IsStatic = h.IsStatic
 	a.ListenPort = h.ListenPort
-	a.LocalListenPort = h.LocalListenPort
 	a.LocalRange = h.LocalRange.String()
 	if isEmptyAddr(a.LocalRange) {
 		a.LocalRange = ""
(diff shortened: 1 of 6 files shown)


models/api_host.go
```diff
@@ -89,7 +88,7 @@ func (a *ApiHost) ConvertAPIHostToNMHost(currentHost *Host) *Host {
 	h.IsK8S = currentHost.IsK8S
 	h.IsStatic = a.IsStatic
 	h.ListenPort = a.ListenPort
-	h.LocalListenPort = currentHost.ListenPort
+	h.ProxyListenPort = a.ProxyListenPort
```
0xdcarns, line 91: Will need to account for public port, correct? Otherwise it may get overwritten to 0

0xdcarns (changes requested):

models/api_host.go
abhishek9686: there is no need to add public port to API model

0xdcarns (approved):

theguy951357 (approved):
built server and two nodes off of branches
joined network "short"
pings ok
in netclient.yml listenport is set to 51821 and proxy and public are set to 51722
changing listen port on the host to 51828
netclient.yml has updated listenport to 51828
pass

PR #1953, opened 2023-01-19
remove wait

afeiszli:
Reason: wait.sh does not work in some environments. In Vultr, it simply hangs forever. Does not seem to support curl or wget. Not sure why, but MQ can just restart until Netmaker starts up, no reason for wait script.


0xdcarns (approved):

theguy951357 (approved):
pass

PR #1958, opened 2023-01-20
decrement access key on use

alphadose:


Aceix (approved):

abhishek9686 (approved):

theguy951357 (approved):
made server and nodes
made network "netty"
made two access keys. one with 999 uses and one with 1 use
999 uses after joining 2 nodes is now at 997
1 use only allows one join
pass

PR #1962, opened 2023-01-23
Gra 1024/mq dyn sec refactor

abhishek9686:
-> Removed multiple roles for mq clients
-> using a generic role for all clients except for the admin client

Testing Steps -
- [x] check server and clients are able to connect to mq
- [x]  check clients and server are able to receive updates and publish messages through mq

### DIFF
--- a/logic/nodes.go
+++ b/logic/nodes.go
@@ -256,7 +256,6 @@ func SetNodeDefaults(node *models.Node) {
 		postdown := parentNetwork.DefaultPostDown
 		node.PostDown = postdown
 	}
-	node.Connected = true
 	// == Parent Network settings ==
 
 	// == node defaults if not set by parent ==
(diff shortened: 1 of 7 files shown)


logic/nodes.go
```diff
@@ -256,6 +256,7 @@ func SetNodeDefaults(node *models.Node) {
 		postdown := parentNetwork.DefaultPostDown
 		node.PostDown = postdown
 	}
+	node.Connected = true
```
0xdcarns, line 259: this is not a good way to handle this also doesn't affect mq, should be discussed and moved to other PR

0xdcarns (changes requested):

mattkasun (approved):

0xdcarns (approved):

theguy951357 (approved):
checkmarks pass

PR #1963, opened 2023-01-23
Remove hosts from relays before deletion

alphadose:


abhishek9686 (approved):

controllers/hosts.go
```diff
@@ -150,7 +150,30 @@ func deleteHost(w http.ResponseWriter, r *http.Request) {
 		logic.ReturnErrorResponse(w, r, logic.FormatError(err, "internal"))
 		return
 	}
-
+	if currHost.IsRelay {
+		if _, _, err := logic.DeleteHostRelay(hostid); err != nil {
+			logger.Log(0, r.Header.Get("user"), "failed to dissociate host from relays:", err.Error())
```
0xdcarns, line 155: need to return error response (http response on all err handles)

0xdcarns (changes requested):

theguy951357:
will wait on changes

controllers/hosts.go
alphadose: done

0xdcarns (approved):

mattkasun (approved):

theguy951357 (approved):
built server from branch

joined two nodes

relayed one node

deleted relaying node

relaying status is gone

repeated but deleted relayed node instead

relayed status is gone

pass

PR #1967, opened 2023-01-24
Move nmctl relay logic to hosts

alphadose:


abhishek9686 (approved):

theguy951357 (approved):
nmctl has relay logic on host section and not on node section
pass

Aceix (approved):

PR #1968, opened 2023-01-24
GRA-1030: update host name on join

abhishek9686:
refer PR https://github.com/gravitl/netclient/pull/156 for testing steps

### DIFF
--- a/controllers/node.go
+++ b/controllers/node.go
@@ -593,7 +593,6 @@ func createNode(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 			logic.UpdateHostFromClient(&data.Host, host) // update the in memory struct values
-			logger.Log(0, fmt.Sprintf("HOST ----------> %+v\n", data.Host))
 			err = logic.UpsertHost(host)
 			if err != nil {
 				logger.Log(0, r.Header.Get("user"),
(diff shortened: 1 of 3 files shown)


theguy951357 (approved):
built server and nodes from netmaker and netclient branches.
joined network setting --name
netclient.yaml has new name in there
host on ui has new name
pass

controllers/node.go
```diff
@@ -592,8 +592,15 @@ func createNode(w http.ResponseWriter, r *http.Request) {
 				logic.ReturnErrorResponse(w, r, logic.FormatError(err, "internal"))
 				return
 			}
-			logic.UpdateHost(&data.Host, host) // update the in memory struct values
-
+			logic.UpdateHostFromClient(&data.Host, host) // update the in memory struct values
+			logger.Log(0, fmt.Sprintf("HOST ----------> %+v\n", data.Host))
+			err = logic.UpsertHost(host)
```
mattkasun, line 597: this is going to overwrite the encoded password and cause issues going forward.

mattkasun (changes requested):

controllers/node.go
abhishek9686: no it won't

0xdcarns: he is maintaining the current host in db

0xdcarns (approved):

PR #1969, opened 2023-01-24
GRA-1045: set node defaults

abhishek9686:
Testing Steps:

- [x] On Adding a host to a network in UI, check if the host is connected to the network in UI
- [x] On joining from netclient to a network check if node status is healthy

### DIFF
--- a/models/node.go
+++ b/models/node.go
@@ -196,13 +196,8 @@ func (node *Node) PrimaryAddress() string {
 }
 
 // Node.SetDefaultConnected
-func (node *LegacyNode) SetDefaultConnected() {
-	if node.Connected == "" {
-		node.Connected = "yes"
-	}
-	if node.IsServer == "yes" {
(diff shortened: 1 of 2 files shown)


models/node.go
```diff
@@ -196,13 +196,8 @@ func (node *Node) PrimaryAddress() string {
...
 
 // Node.SetDefaultConnected
-func (node *LegacyNode) SetDefaultConnected() {
-	if node.Connected == "" {
-		node.Connected = "yes"
-	}
-	if node.IsServer == "yes" {
-		node.Connected = "yes"
-	}
+func (node *Node) SetDefaultConnected() {
```
mattkasun, line 199: do we really need a function for this?

alphadose (approved):

Aceix (approved):

theguy951357 (approved):
checkmarks pass

PR #1971, opened 2023-01-25
Fix address range check

alphadose:


Aceix (approved):

0xdcarns (approved):

theguy951357 (commented):
still getting errors when changing an ipv6 address. then netclient list shows an ipv6 address of <nil> and then gets a 401 unauthorized error when leaving the network

mattkasun (approved):

theguy951357 (approved):
can update the ipv6 on node through nmctl
pass

PR #1972, opened 2023-01-25
Fix nmctl host update

alphadose:


Aceix (approved):

0xdcarns (approved):

theguy951357 (approved):
built nmctl
ran nmctl host update <host_id> --default
host is now default host on the UI
pass

PR #1982, opened 2023-01-27
fix node ipv6 address update

alphadose:


Aceix (approved):

theguy951357 (approved):
dual stack updates successfully
pass

PR #1983, opened 2023-01-27
Fix host password update

alphadose:


Aceix (approved):

logic/hosts.go
```diff
@@ -160,6 +160,9 @@ func UpdateHostFromClient(newHost, currHost *models.Host) (sendPeerUpdate bool)
 		currHost.EndpointIP = newHost.EndpointIP
 		sendPeerUpdate = true
 	}
+	if newHostPass, err := bcrypt.GenerateFromPassword([]byte(newHost.HostPass), 5); err == nil && string(newHostPass) != currHost.HostPass {
+		currHost.HostPass = string(newHostPass)
+	}
```
abhishek9686, line 165: but why would the password get updated for a host?

alphadose: cc @mattkasun

mattkasun: I can't think of a use case for updating the Host password.  If it is required, this implementation is not complete in that the mq dynamic-security is not being updated.

mattkasun (changes requested):
I don't think we should be updating the host password.

theguy951357:
will wait on changes

PR #1984, opened 2023-01-27
Add expired nodes to zombie list and send peer update on deletion

alphadose:


abhishek9686 (commented):
can't node expiration be handled during check-in?

alphadose:
the `CheckZombies` function appends the node id to a slice which gets cleaned up all together at once by the `ManageZombies` goroutine

abhishek9686 (approved):
can't node expiration be handled during check-in?

logic/zombie.go
```diff
@@ -80,11 +81,13 @@ func ManageZombies(ctx context.Context) {
...
 						continue
 					}
-					if time.Since(node.LastCheckIn) > time.Minute*ZOMBIE_DELETE_TIME {
+					if time.Since(node.LastCheckIn) > time.Minute*ZOMBIE_DELETE_TIME || time.Now().After(node.ExpirationDateTime) {
 						if err := DeleteNode(&node, true); err != nil {
 							logger.Log(1, "error deleting zombie node", zombies[i].String(), err.Error())
 							continue
 						}
+						node.Action = models.NODE_DELETE
+						peerUpdate <- &node
```
mattkasun, line 90: why use a channel here; why not just call mq.NodeUpdate ?

main.go
```diff
@@ -191,7 +191,15 @@ func runMessageQueue(wg *sync.WaitGroup) {
 	mq.SetupMQTT()
 	ctx, cancel := context.WithCancel(context.Background())
 	go mq.Keepalive(ctx)
-	go logic.ManageZombies(ctx)
+	go func() {
```
mattkasun, line 194: why do we need this go routine.

mattkasun (commented):
questions

theguy951357:
will wait on changes

logic/zombie.go
abhishek9686: i guess it's because of cyclic import calling `mq` from logic pkg

alphadose: @abhishek9686 yes

main.go
alphadose: for sending peer update on node deletion by `ManageZombies`. main package for no cyclic dependency

theguy951357:
I set the node to the year 2000 and watch the logs. I have not seen the expired node go into the zombie list.

theguy951357 (approved):
expired node went away
pass

PR #1985, opened 2023-01-27
skip if ext client doesn't belong to ingress GW

abhishek9686:
Testing steps:

-> create two ingress gateways 
-> create ext clients on both the ingress nodes
-> verify whether extpeers are only added to their respective ingress node
-> verify on other peers in the network, whether extclient client allowed IP is appended to the right ingress node


0xdcarns (approved):

theguy951357 (approved):
joined three nodes.
made two nodes an ingress.
added an ext client to each node.
wg on all three nodes show that each client is connected to its assigned node.
pass

PR #2008, opened 2023-02-09
disallowed updating of egress ranges

0xdcarns:
To test:

- [x] use a UI that enables updating of egress ranges and ensure it doesn't actually update when you refresh for an egress gateway


mattkasun (approved):

theguy951357 (approved):
egress range does not update

PR #2010, opened 2023-02-10
Gra 1153 database sync

0xdcarns:
- [x] ensure API updates, reads and deletes function as normal.
- [x] bonus test: the scale test and see if it makes a difference


mattkasun (approved):

theguy951357 (approved):
pass

PR #2013, opened 2023-02-10
added logic to add default hosts to networks upon network creation

0xdcarns:
- [x] make a host default
- [x] create a new network
- [x] check if host makes a node on that network


theguy951357 (approved):
checkmarks pass

PR #2015, opened 2023-02-10
Upgrade Script Updates

afeiszli:


afeiszli:
- nmctl is currently not working correctly, blocking completion
-  also need to implement Egress

afeiszli:
Upgrade script now functional. Test Steps
1. Create a 0.17.1 server with nm-quick
2. Create a few networks
3. Add a node besides server node
4. Set server node as an Ingress, Egress, and Relay in various configurations
5. Run upgrade script
6. See that node is created on host and retains settings
7. Re-run this test with Netmaker EE

LIMITATIONS:
1. Cannot currently retain private IP address due to bug in nmctl (GRA-1256)
2. Cannot retain a Relay node due to move from node to host (server is getting updated before all nodes, so there are no hosts for it to relay)
3. Cannot retain ext clients

afeiszli:
Note: LATEST tag is on 0.18.1 for testing. Won't work until release is out for current version. Will need to be changed to actual latest upon release.

afeiszli:
@theguy951357 readt for retest

theguy951357 (approved):
server upgrades with new client replacement for netmaker-1
pass

0xdcarns (approved):

PR #2024, opened 2023-02-13
GRA-1199: add routes for all ext clients in the network

abhishek9686:
Testing:

- [x] check if extclients on two different ingress GW are able to talk with each other


mattkasun (approved):

theguy951357:
joined two nodes, made an ingress for each node. made ext client for each ingress. I am getting request timeouts when pinging from one client to the other. 

Screenshot 2023-02-14 at 10 58 27 AM

theguy951357 (approved):
two ext clients can ping eachother from different ingress nodes
pass

PR #2037, opened 2023-02-14
Fix issue with deleting nodes from default hosts

0xdcarns:
To test: Use client in PR https://github.com/gravitl/netclient/pull/218

- [x] Ensure a default Host gets via Nodes added to all networks created after initial join
- [x] Ensure user can delete those nodes
- [x] Ensure user can delete the host after the nodes


mattkasun (approved):

Aceix (approved):

theguy951357 (approved):
checkmarks pass

PR #2043, opened 2023-02-16
Various test enhancments, fix go vet, update CI

yunginnanet:
Passes go vet now and adds additional error checks in tests :)

### DIFF
--- a/controllers/user_test.go
+++ b/controllers/user_test.go
@@ -3,20 +3,24 @@ package controller
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/gravitl/netmaker/logic"
 	"github.com/gravitl/netmaker/models"
-	"github.com/stretchr/testify/assert"
 )
 
-func deleteAllUsers() {
+func deleteAllUsers(t *testing.T) {
+	t.Helper()
 	users, _ := logic.GetUsers()
 	for _, user := range users {
-		logic.DeleteUser(user.UserName)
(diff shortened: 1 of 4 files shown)


controllers/user_test.go
```diff
@@ -149,63 +155,76 @@ func TestValidateUser(t *testing.T) {
...
-		logic.CreateUser(&user)
+		if err := logic.CreateUser(&user); err != nil {
+			t.Error(err)
+		}
 		admins, err := logic.GetUsers()
 		assert.Nil(t, err)
 		for _, u := range admins {
 			if u.UserName == "admin" {
-				assert.Equal(t, "admin", u.UserName)
+				assert.Equal(t, true, u.IsAdmin)
```
yunginnanet, line 204: Note that this was changed because in this test you we were asserting something to be true that had no possibility of being false

netclient/ncutils/netclientutils.go
```diff
@@ -166,13 +166,16 @@ func GetPublicIP(api string) (string, error) {
...
 			var bodyBytes []byte
 			bodyBytes, err = io.ReadAll(resp.Body)
 			if err != nil {
+				if resp.Body != nil {
+					_ = resp.Body.Close()
+				}
 				continue
 			}
+			_ = resp.Body.Close()
 			endpoint = string(bodyBytes)
```
yunginnanet, line 179: If you use defer in a for loop, the bodies don't get Closed until after the function is done, this creates a pretty nasty build-up of open response bodies as the for loop continues.

mattkasun (approved):

PR #2064, opened 2023-02-23
Gra 1264 better peer mgmt

0xdcarns:
## Describe your changes
Removed some useless code. Fixed handling deleting peers better. Pairs with netclient story/GRA-1264-better-peer-mgmt

## Provide testing steps
- [x] Install netclient on a few hosts
- [x] Join multiple networks with 3+ peers in each
- [x] Upon joining ensure WG connections were formed
- [x] Delete node from one and ensure allowed ips are updated
- [x] Delete the same completely and ensure peer is removed quickly while not affecting other connections

- [x] I have had my code reviewed by a peer.
- [x] Netmaker is awesome.

### DIFF
--- a/logic/peers.go
+++ b/logic/peers.go
@@ -41,7 +41,7 @@ func GetProxyUpdateForHost(host *models.Host) (models.ProxyManagerPayload, error
 		relayPeersMap := make(map[string]models.RelayedConf)
 		for _, relayedHost := range relayedHosts {
 			relayedHost := relayedHost
-			payload, err := GetPeerUpdateForHost("", &relayedHost)
+			payload, err := GetPeerUpdateForHost("", &relayedHost, nil)
 			if err == nil {
 				relayedEndpoint, udpErr := net.ResolveUDPAddr("udp", fmt.Sprintf("%s:%d", relayedHost.EndpointIP, GetPeerListenPort(&relayedHost)))
 				if udpErr == nil {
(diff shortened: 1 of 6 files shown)


logic/peers.go
```diff
@@ -514,9 +352,32 @@ func GetPeerUpdateForHost(network string, host *models.Host) (models.HostPeerUpd
 		}
 	}
 
+	// run through delete nodes
+	if len(deletedNodes) > 0 {
```
mattkasun, line 356: unnecessary check as for loop will be skipped if deletedNodes is empty.  The compiler will probably optimize out the check so not a big deal

mattkasun (approved):

theguy951357 (approved):
all check marks pass
nodes maintained connectivity throughout.
pass

logic/peers.go
```diff
@@ -328,10 +159,14 @@ func GetPeerUpdateForHost(network string, host *models.Host) (models.HostPeerUpd
...
 		for _, peer := range currentPeers {
+			peer := peer
 			if peer.ID == node.ID {
 				logger.Log(2, "peer update, skipping self")
 				//skip yourself
-
+				continue
+			}
+			if peer.Action == models.NODE_DELETE || peer.PendingDelete {
+				deletedNodes = append(deletedNodes, peer) // track deleted node for peer update
```
abhishek9686, line 169: wouldn't this approach only add zombie nodes to deleted nodes list, and peers which are deleted already would never be removed on client

0xdcarns: the zombie check timeout was extended to discount for this

PR #2069, opened 2023-02-24
GRA-813 Better use of context in main

0xdcarns:
## Describe your changes
Fixed so that all main go routines share the same context.

## Provide Issue ticket number if applicable/not in title
title.

## Provide testing steps
- [x] Start a server
- [x] Be able to run (when connected to broker)
- [x] Be able to call docker stop
- [x] Start server locally with MESSAGEQUEUE_BACKEND=off and be able to ctrl+c easily

- [x] I have had my code reviewed by a peer.
- [x] Netmaker is awesome.

### DIFF
--- a/controllers/controller.go
+++ b/controllers/controller.go
@@ -4,11 +4,8 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"os"
-	"os/signal"
 	"strings"
 	"sync"
-	"syscall"
 	"time"
 
 	"github.com/gorilla/handlers"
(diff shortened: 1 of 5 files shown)


theguy951357 (approved):
checkmarks pass
pr checklist passes

mattkasun (changes requested):
From the documentation of context

Programs that use Contexts should follow these rules to keep interfaces consistent across packages and enable static analysis tools to check context propagation:


Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

0xdcarns:
> From the documentation of context
> 
> Programs that use Contexts should follow these rules to keep interfaces consistent across packages and enable static analysis tools to check context propagation:
> 
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

cool that's not the point of this ticket, I was just adapting current code, please review the ticket

afeiszli (approved):

PR #2158, opened 2023-03-30
add checks to user update processing

mattkasun:
## Describe your changes
also addresses GRA-1480
- [ ] Netmaker is awesome.

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -331,7 +331,7 @@ func updateUser(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	var params = mux.Vars(r)
 	// start here
-	jwtUser, _, isadmin, err := logic.VerifyJWS(r.Header.Get("Authorization"))
+	jwtUser, _, isadmin, err := logic.VerifyJWT(r.Header.Get("Authorization"))
 	if err != nil {
 		logger.Log(0, "verifyJWT error", err.Error())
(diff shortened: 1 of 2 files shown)


logic/jwts.go
```diff
@@ -101,6 +102,18 @@ func CreateUserJWT(username string, networks []string, isadmin bool) (response s
 	return "", err
 }
 
+// VerifyJWT verifies Auth Header
```
0xdcarns, line 105: either comment or func name is incorrect

0xdcarns (changes requested):

logic/jwts.go
mattkasun: updated func name

controllers/user.go
```diff
@@ -331,7 +331,18 @@ func updateUser(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	var params = mux.Vars(r)
 	// start here
+	jwtUser, _, isadmin, err := logic.VerifyJWT(r.Header.Get("Authorization"))
+	if err != nil {
```
abhishek9686, line 335: can't permissions be verified at middleware?

theguy951357:
will wait on changes

controllers/user.go
mattkasun: don't see the point of creating new middleware for a single endpoint.

theguy951357 (approved):
non-admin user cannot change the credentials of an admin user.
critical list passes

0xdcarns (approved):

PR #2241, opened 2023-04-26
GRA-1494: Turn serverside Implementation

abhishek9686:
## Provide testing steps
 refer to https://github.com/gravitl/netclient/pull/371 for steps
@theguy951357 the docker-compose needs to be updated to test this pr, please check for the changes
- [x] Netmaker is awesome.


mq/handlers.go
```diff
@@ -163,6 +163,11 @@ func UpdateHost(client mqtt.Client, msg mqtt.Message) {
 			return
 		}
 		sendPeerUpdate = true
+	case models.RegisterWithTurn:
+		err = logic.RegisterHostWithTurn(hostUpdate.Host.ID.String(), hostUpdate.Host.HostPass)
+		if err != nil {
+			logger.Log(0, "failed to register host with turn server: ", err.Error())
+		}
```
mattkasun, line 170: missing return?

turnserver/config/config.go
```diff
@@ -0,0 +1,99 @@
+package config
```
mattkasun, line 1: is it necessary to have all these sub-package especially as most have but a single file.

turnserver/src/controller/controller.go
```diff
@@ -0,0 +1,80 @@
+package controller
```
mattkasun, line 1: is there a reason for the extra src dir?

turnserver/internal/utils/response.go
```diff
@@ -0,0 +1,35 @@
+package utils
```
mattkasun, line 1: these functions should be in the src/controller pkg.. For that matter, could be added to controller.go

turnserver/config/config.go
```diff
@@ -0,0 +1,99 @@
...
+// GetAllowedOrigin - get the allowed origin
+func GetAllowedOrigin() string {
+	allowedorigin := "*"
+	if os.Getenv("CORS_ALLOWED_ORIGIN") != "" {
+		allowedorigin = os.Getenv("CORS_ALLOWED_ORIGIN")
+	}
+	return allowedorigin
+}
+
+// SetVersion - set version of netmaker
```
mattkasun, line 21: This function is not used

turnserver/config/config.go
mattkasun, line 1: if main was built as a corba app; could get rid of most of these functions.... define flags for the vars and then they can be set using the flag or env vars... this would make it easier to deploy are bare-metal should someone decide to use bare-metal instead of containers

turnserver/internal/auth/auth.go
```diff
@@ -0,0 +1,65 @@
...
+
+var (
+	authMapLock    = &sync.RWMutex{}
+	HostMap        = make(map[string]string)
+	authBackUpFile = "auth.json"
+	backUpFilePath = filepath.Join("/etc/config", authBackUpFile)
+)
+
+func init() {
+	os.MkdirAll("/etc/config", os.ModePerm)
```
mattkasun, line 23: should this path be /etc/turn or /etc/turn/config?

turnserver/internal/auth/auth.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/gravitl/netmaker/logger"
+	"github.com/gravitl/netmaker/turnserver/config"
+	"github.com/pion/turn/v2"
+)
+
+var (
+	authMapLock    = &sync.RWMutex{}
+	HostMap        = make(map[string]string)
+	authBackUpFile = "auth.json"
+	backUpFilePath = filepath.Join("/etc/config", authBackUpFile)
```
mattkasun, line 19: missing slash?  /etc/config/
see next comment re: path

turnserver/src/middleware/middleware.go
```diff
@@ -0,0 +1,25 @@
+package middleware
```
mattkasun, line 1: couldn't this be included in controller.go

turnserver/src/routes/routes.go
```diff
@@ -0,0 +1,34 @@
...
+	r.DELETE("/host/deregister", gin.BasicAuth(gin.Accounts{
+		config.GetUserName(): config.GetPassword(),
+	}), host.Remove)
+	r.GET("/status", gin.BasicAuth(gin.Accounts{
+		config.GetUserName(): config.GetPassword(),
+	}), status)
+}
+
+func status(c *gin.Context) {
+	c.JSON(http.StatusOK, struct {
```
mattkasun, line 31: this would be clearer by using gin.H
c.JSON(http.StatusOK, gin.H{"msg":"active"}

mattkasun (changes requested):
Is the netmaker repo the best place to turn or should it be in its own repo?

abhishek9686:
Will open a different PR, after moving turn to a separate repo

turnserver/src/controller/controller.go
abhishek9686: just for code structure

turnserver/config/config.go
abhishek9686: agreed, can be put in the backlog I don't see it as an immediate required change

turnserver/internal/auth/auth.go
abhishek9686: filepath.join will add slash for comma separated values
yeah we could put it in /etc/turn/config for more isolation maybe

PR #2325, opened 2023-05-22
NET-164: Refactor Peer updates - Ext. Clients

abhishek9686:
## Describe your changes

1. Refactored Ext. Client peer updates

## Provide testing steps

- [x] create/update/delete an ext client, verify whether updates are going correctly to peers.
- [x] netclient should be built from https://github.com/gravitl/netclient/pull/407 

- [x] Netmaker is awesome.


auth/host_session.go
```diff
@@ -237,16 +237,17 @@ func CheckNetRegAndHostUpdate(networks []string, h *models.Host) {
 				Host:   *h,
 				Node:   *newNode,
 			})
+			if servercfg.IsMessageQueueBackend() {
```
afeiszli, line 240: Why do we have this check? Shouldn't we always assume it's a message queue? Otherwise just wouldn't work, right?

abhishek9686: this check is at multiple places, would be better to get rid of it in a separate PR.
I believe we have a ticket already to do this.

afeiszli (approved):

theguy951357 (approved):
pr steps pass. CRUD operations work on ext client.
critical list skipped until refactor_peer_updates is complete

Aceix (approved):

PR #2332, opened 2023-05-24
NET-246: Use the default config as a fallback

TobiaszCudnik:
## Describe your changes

Changes:
- explicitly fall back to a default config value when the value is empty
- separate `netmaker.default.env` (with comments)
- fixed `TURN_PASSWORD`
- added versioning to nm-quick

This PR is a no-release hotfix and should be cherrypicked to `master`. 

## Provide testing steps

```
wget https://raw.githubusercontent.com/gravitl/netmaker/NET-246/read-def-config/scripts/nm-quick.sh
chmod +x nm-quick.sh
./nm-quick.sh -b local -t NET-246/read-def-config
```

- [x] Netmaker is awesome.


scripts/nm-quick.sh
```diff
@@ -734,8 +737,7 @@ install_netmaker() {
 			wget -qO "$SCRIPT_DIR"/docker-compose.override.yml $COMPOSE_OVERRIDE_URL
 		fi
 		wget -qO "$SCRIPT_DIR"/Caddyfile "$CADDY_URL"
-		wget -qO "$SCRIPT_DIR"/netmaker.env "$BASE_URL/scripts/netmaker.env"
-		ln -fs "$SCRIPT_DIR/netmaker.env" "$SCRIPT_DIR/.env"
+		wget -qO "$SCRIPT_DIR"/netmaker.default.env "$BASE_URL/scripts/netmaker.default.env"
```
mattkasun, line 740: why not just save this as .env instead of creating symbolic link below

mattkasun (changes requested):

scripts/nm-quick.sh
TobiaszCudnik: User editable config has a human readable name and isnt a hidden file.

1. `netmaker.env` - main config, user editable
2. `netmaker.default.env` - defaults shipped with every release
3. `.env` - consumed by machines

mattkasun: disagree, but ok.... but you are saving it as netmaker.default.env instead of netmaker.env so the symlink command below will fail

mattkasun (changes requested):

scripts/nm-quick.sh
TobiaszCudnik: > disagree, but ok....

Any reasons?

> you are saving it as netmaker.default.env instead of netmaker.env so the symlink command below will fail

```
$ ln -s missing foo
$ ll
lrwxrwxrwx 1 root root    7 May 24 10:26 foo -> missing
```

afeiszli (approved):

PR #2395, opened 2023-06-12
prevent ingress creation on NATed host

mattkasun:


Aceix (approved):

logic/gateway.go
```diff
@@ -112,6 +112,9 @@ func CreateIngressGateway(netid string, nodeid string, ingress models.IngressReq
 	if host.FirewallInUse == models.FIREWALL_NONE {
 		return models.Node{}, errors.New("firewall is not supported for ingress gateways")
 	}
+	if host.NatType != models.NAT_Types.Public {
+		return models.Node{}, errors.New("ingress cannot be created on nodes behind NAT")
+	}
```
abhishek9686, line 117: why not allow it?
clients can still connect over the public listen port right?

abhishek9686 (changes requested):

logic/gateway.go
TobiaszCudnik: Whats a "public listen port" if not a port **not** behind NAT (called "public NAT" in this case).

afeiszli: mentioned by @TobiaszCudnik if there is port forwarding, it should be detected as public anyway

afeiszli (approved):

theguy951357 (approved):
pr passes
critical list passses

PR #2399, opened 2023-06-16
NET-344 NET-341 DB caching

TobiaszCudnik:
## Describe your changes

Fixes NET-344 NET-341

Added DB caching

## Checklist before requesting a review
- [ ] My changes affect only 10 files or less.
- [ ] I have performed a self-review of my code and tested it.
- [ ] If it is a new feature, I have added thorough tests, my code is <= 1450 lines.
- [ ] If it is a bugfix, my code is <= 200 lines.
- [ ] My functions are <= 80 lines.
- [ ] I have had my code reviewed by a peer.
- [ ] My unit tests pass locally.
- [ ] Netmaker is awesome.

PR #2404, opened 2023-06-19
NET-345 memleak fix from #2399

TobiaszCudnik:
## Describe your changes

Extract from #2399.

## Provide testing steps

1. Start netmaker
2. Check the memory
3. Wait 24h
4. Check the memory

Result: Should be less with this PR than without it.

- [ ] Netmaker is awesome.

### DIFF
--- a/logic/zombie.go
+++ b/logic/zombie.go
@@ -77,6 +77,10 @@ func checkForZombieHosts(h *models.Host) {
 func ManageZombies(ctx context.Context, peerUpdate chan *models.Node) {
 	logger.Log(2, "Zombie management started")
 	InitializeZombies()
+
+	// run this check 4 times a day
+	duration := time.Hour * ZOMBIE_TIMEOUT
+	delay := time.NewTimer(duration)
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


logic/zombie.go
abhishek9686, line 1: using a ticker would be more appropriate.
reference - https://gobyexample.com/tickers

abhishek9686 (changes requested):

logic/zombie.go
TobiaszCudnik: I agree but theres no benefit in changing it now - it will work the same way, the memleak is the issue of this PR.

abhishek9686: Also i don't quite understand how is `time.After(time.Hour * ZOMBIE_TIMEOUT)` causing mem leak, could you explain?

abhishek9686:
PR should be raised against `develop`, please change the destination branch

logic/zombie.go
mattkasun: > I agree but theres no benefit in changing it now - it will work the same way, the memleak is the issue of this PR.

The timer was added as part of this PR, so the recommendation to use ticker is valid

PR #2461, opened 2023-07-24
[NET-353] Solve race condition with "unique" ips

### LINKED ISSUE #2356: [Bug]: Racing condition when multiple hosts join the same network concurrently
### What happened?

When multiple hosts join the same network concurrently then some end up having the same private IPv4 address and some hosts don't get registered at all. Those with the same IPv4 address have a "Warning" badge in the Health Status column.

### Version

v0.20.0

### What OS are you using?

Linux

### Relevant log output

_No response_

### Contributing guidelines

- [X] Yes, I did.

gabrielseibel1:
## Describe your changes

Makes the procedure of getting-and-setting unique ips to new nodes and extclients atomic and mutually exclusive by using a lock.

Closes #2356 

## Provide testing steps

1. create two or more nodes simultaneously (create a script that calls the api in parallel goroutines?)
2. run this test a few times (the error has a <100% probability of occurring)

- [x] Netmaker is awesome.

### DIFF
--- a/logic/extpeers.go
+++ b/logic/extpeers.go
@@ -150,7 +150,6 @@ func GetExtClientByPubKey(publicKey string, network string) (*models.ExtClient, 
 
 // CreateExtClient - creates an extclient
 func CreateExtClient(extclient *models.ExtClient) error {
-
 	// lock because we need unique IPs and having it concurrent makes parallel calls result in same "unique" IPs
 	addressLock.Lock()
 	defer addressLock.Unlock()
--- a/logic/extpeers.go
+++ b/logic/extpeers.go
@@ -151,6 +151,10 @@ func GetExtClientByPubKey(publicKey string, network string) (*models.ExtClient, 
 // CreateExtClient - creates an extclient
 func CreateExtClient(extclient *models.ExtClient) error {
 
+	// lock because we need unique IPs and having it concurrent makes parallel calls result in same "unique" IPs
+	addressLock.Lock()
+	defer addressLock.Unlock()
+
 	if len(extclient.PublicKey) == 0 {
 		privateKey, err := wgtypes.GeneratePrivateKey()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


logic/networks.go
```diff
@@ -424,3 +425,9 @@ func SortNetworks(unsortedNetworks []models.Network) {
 }
 
 // == Private ==
+
+var addressLock sync.Locker
+
+func init() {
```
mattkasun, line 431: This function is not required if address lock is declared to be &sync.Mutex{}

logic/extpeers.go
```diff
@@ -150,6 +150,9 @@ func GetExtClientByPubKey(publicKey string, network string) (*models.ExtClient,
 
 // CreateExtClient - creates an extclient
 func CreateExtClient(extclient *models.ExtClient) error {
+	// lock because we need unique IPs and having it concurrent makes parallel calls result in same "unique" IPs
+	addressLock.Lock()
```
mattkasun, line 154: this lock is premature.  It should be only around the  functionality that needs mutual exclusion, namely GetUniqueAddress and GetUniqueAddress6

mattkasun (changes requested):

logic/extpeers.go
gabrielseibel1: i believe that approach wouldn't solve the problem that needs addressing, because making the "getting" of the addresses mutually exclusive alone doesn't account for the "setting" part. Getting the unique ip and setting it should be atomic, they can't be separate from each other, otherwise we can still have race conditions. In the case we don't make it atomic, there can be two sequential (synchronized) calls to getting addresses (say, A and B), but if the second one (B) gets the address lock before the first one (A) sets the node/client in the db, then B will have the same "unique" address that A got.

Aceix (approved):

mattkasun (approved):

abhishek9686 (approved):

PR #2471, opened 2023-07-27
Net 487: delete ingress network parameter 

mattkasun:
ignore network parameter passed to delete ingress endpoint
a node can only be part of a single network and calling the endpoint with incorrect network name results in extclients not being deleted.


Aceix (approved):

abhishek9686 (approved):

theguy951357 (approved):
tested and passed

PR #2476, opened 2023-07-28
enforce unique names for ext client names

mattkasun:
prevent creation of client with same name as existing client


Aceix (approved):

Aceix (approved):

abhishek9686:
why are there so many changes related to this ticket?
ideally, there should be only one new function to validate the new name whether it's unique

mattkasun:
> why are there so many changes related to this ticket? ideally, there should be only one new function to validate the new name whether it's unique

There is a modified function isValid that does the check.  The check has to be done not only on creation but also on an update but only if the update modifies the client id.  There was an existing function that had undocumented side-effects and logs where updated to use slog if the logs were in an area where the code was being changed.l

PR #2495, opened 2023-08-03
[NET-494 / ACC-322] New free tier limits

gabrielseibel1:
## Describe your changes

Introduces new free tier limits (machines, ingresses, egresses)

NEW env vars:

- MACHINES_LIMIT
- INGRESSES_LIMIT
- EGRESSES_LIMIT

New metrics:
- Ingresses
- Egresses


abhishek9686 (changes requested):

logic/hosts.go
```diff
@@ -162,7 +162,7 @@ func CreateHost(h *models.Host) error {
 	if err != nil && !database.IsEmptyRecord(err) {
 		return err
 	}
-	if len(hosts) >= Hosts_Limit {
+	if len(hosts) >= HostsLimit {
```
abhishek9686, line 165: Shouldn't this be `MachinesLimit`?

abhishek9686 (changes requested):

logic/gateway.go
```diff
@@ -28,6 +57,13 @@ func CreateEgressGateway(gateway models.EgressGatewayRequest) (models.Node, erro
 		return models.Node{}, errors.New("firewall is not supported for egress gateways")
 	}
 	for i := len(gateway.Ranges) - 1; i >= 0; i-- {
+		// check if internet gateway IPv4
+		if gateway.Ranges[i] == "0.0.0.0/0" && FreeTier {
+			logger.Log(0, "currently IPv4 internet gateways are not supported on the free tier", gateway.Ranges[i])
```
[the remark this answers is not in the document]
gabrielseibel1: makes sense, will do that

logic/hosts.go
gabrielseibel1: i think you're correct, we should be checking for the machines limit here. That said, I think, to be flexible, we should check for hosts limit too. Perhaps the limit will be maxint, but what if we decide to change the env var to say 2? Then we'll already have the code implemented to limit it

abhishek9686:
Also you can deprecate CLIENT_LIMIT and HOST_LIMIT

logic/hosts.go
gabrielseibel1: what do you think?

gabrielseibel1:
We could, but then it would be cumbersome to introduce it back, and we may do that. I think we should have all the limitation mechanisms in place and then just control the limits via env vars, and they can be a large value instead of being deprecated - to be flexible

logic/hosts.go
abhishek9686: The reason machines limit was introduced was too limit both clients and hosts creation after limit has been reached on free tier

abhishek9686:
> We could, but then it would be cumbersome to introduce it back, and we may do that. I think we should have all the limitation mechanisms in place and then just control the limits via env vars, and they can be a large value instead of being deprecated - to be flexible

They are not supposed to be some arbitrary large number, it should be unused in the code since machine limit is introduced in the code

PR #2498, opened 2023-08-07
fix(NET-486): change client name length validation

Aceix:
set limit to 5<=x<=32

- [ ] Netmaker is awesome.

### DIFF
--- a/controllers/regex_test.go
+++ b/controllers/regex_test.go
@@ -27,18 +27,32 @@ func TestValidName(t *testing.T) {
 			Want: false,
 		},
 		{
-			Name: "nametoolong",
+			Name: "longname",
 			Args: args{
 				Name: "TestvalidNameTestvalidName",
 			},
-			Want: false,
+			Want: true,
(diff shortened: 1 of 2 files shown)


controllers/regex_test.go
```diff
@@ -27,19 +27,33 @@ func TestValidName(t *testing.T) {
...
+			Name: "longname",
 			Args: args{
 				Name: "TestvalidNameTestvalidName",
 			},
-			Want: false,
+			Want: true,
 		},
 		{
-			Name: "maxlength",
+			Name: "max length",
```
abhishek9686, line 37: are spaces allowed in the client name?

abhishek9686 (changes requested):

controllers/regex_test.go
Aceix: this is the test name rather. and no, spaces are not allowed

abhishek9686 (approved):

PR #2505, opened 2023-08-08
set endpoint and port when registering

Aceix:
## Describe your changes

check NET-497

- [ ] Netmaker is awesome.

### DIFF
--- a/auth/host_session.go
+++ b/auth/host_session.go
@@ -15,6 +15,8 @@ import (
 	"github.com/gravitl/netmaker/models"
 	"github.com/gravitl/netmaker/mq"
 	"github.com/gravitl/netmaker/servercfg"
+	"golang.org/x/exp/slices"
+	"golang.org/x/exp/slog"
 )
 
 // SessionHandler - called by the HTTP router when user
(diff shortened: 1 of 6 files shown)


auth/host_session.go
```diff
@@ -140,7 +142,22 @@ func SessionHandler(conn *websocket.Conn) {
 					return
 				}
 			}
-			logic.CheckHostPorts(&result.Host)
+			if slices.Contains(registerMessage.CustomParams, models.HostRegisterCustomParamListenPort) {
```
abhishek9686, line 145: this fields are already on host model, not required to extend the req to have custom params

auth/host_session.go
abhishek9686, line 1: I don't think these changes are required, all the changes should be only on client side for user to set IP and Port for a host

abhishek9686 (changes requested):

auth/host_session.go
Aceix: Mainly had to do this for the port check. Otherwise `logic.CheckHostPorts(&result.Host)` will automatically assign any free port even when user specifies one

PR #2506, opened 2023-08-08
NET-507

uGiFarukh:
## Describe your changes
An expired node now gets cleaned up every one hour after routine checkup. Expired node time now do not get reset to default on server restart.

## Provide testing steps
1. join a client to a network
2. edit the node and change the expiration date to a minute in the future
3. wait approximately 1 hour
4. expired nodes cleanup routine should delete the node
5. Repeat steps 1 and 2
6. Restart the netmaker server
7. Verify if the expiry date and time on the node is same as previously set

- [ ] Netmaker is awesome.

### DIFF
--- a/logic/nodes.go
+++ b/logic/nodes.go
@@ -1,6 +1,7 @@
 package logic
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -20,6 +21,7 @@ import (
 	"github.com/gravitl/netmaker/models"
 	"github.com/gravitl/netmaker/servercfg"
 	"github.com/gravitl/netmaker/validation"
+	"golang.org/x/exp/slog"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


mattkasun (approved):

logic/nodes.go
```diff
@@ -444,6 +446,35 @@ func GetAllNodesAPI(nodes []models.Node) []models.ApiNode {
 	return apiNodes[:]
 }
 
+// DeleteExpiredNodes - goroutine which deletes nodes which are expired
+func DeleteExpiredNodes(ctx context.Context, peerUpdate chan *models.Node) {
+	for {
+		select {
+		case <-ctx.Done():
+			close(peerUpdate)
```
abhishek9686, line 454: we are already closing channel here, the program will panic closing channel twice
https://github.com/gravitl/netmaker/blob/d978cfe0c896760dab65d9999cc06c30181416cd/logic/zombie.go#L83

logic/nodes.go
```diff
@@ -444,6 +446,35 @@ func GetAllNodesAPI(nodes []models.Node) []models.ApiNode {
...
+		case <-time.After(time.Hour):
+			// Delete Expired Nodes Every Hour
+			allnodes, err := GetAllNodes()
+			if err != nil {
+				slog.Error("failed to retrieve all nodes", "error", err.Error())
+				return
+			}
+			for _, node := range allnodes {
+				if time.Now().After(node.ExpirationDateTime) {
+					if err := DeleteNode(&node, true); err != nil {
```
abhishek9686, line 465: the purge should be `false` here since the actual delete happens when the node reacts to the node delete action

logic/nodes.go
```diff
@@ -444,6 +446,35 @@ func GetAllNodesAPI(nodes []models.Node) []models.ApiNode {
...
+						slog.Error("error deleting expired node", "nodeid", node.ID.String(), "error", err.Error())
+						continue
+					}
+					node.Action = models.NODE_DELETE
+					peerUpdate <- &node
```
abhishek9686, line 470: also set here node.PendingDelete = true

abhishek9686 (changes requested):

abhishek9686 (approved):

PR #2538, opened 2023-08-23
[NET-546] Move ee code to ee package, unify ee status and terminology

gabrielseibel1:
## Describe your changes

Moved all pieces of code that where specific to EE/Pro to correct directory.

Used variable functions that are filled in with pro code on pro initialization.

Main areas affected are metrics, relays, and ext. client acls.

Lot of the refactoring was done with automation/IDE tools, and go vet, staticcheck and go test pass, but nonetheless we should manually test this thoroughly since the diff is huge... lots of moving this around but logic should be the same.

Consolidated naming for paid/professional/enterprise version as "pro". Notes:
- Changes image tags
- Changes build tags
- Changes package names
- Doesn't change links to docs that mention "ee"
- Doesn't change parameters sent to PostHog that mention "ee"

## Provide testing steps

Test that pro features mentioned above are working and not available on community


abhishek9686 (changes requested):

ee/initialize.go
abhishek9686, line 1: let's stick to only one terminology
should change EE to PRO everywhere in the code

abhishek9686 (changes requested):

ee/initialize.go
gabrielseibel1: https://github.com/gravitl/netmaker/pull/2538/commits/6f6c2aa6e230c54a93b1a70cd51961f3bbd1cfd2

abhishek9686 (approved):

mattkasun (approved):

abhishek9686 (changes requested):
@gabrielseibel1 can you revert changes of ee -> pro for docker images and workflow, tags

pro/initialize.go
```diff
@@ -1,33 +1,33 @@
 //go:build ee
```
Aceix, line 1: may need to change this to pro as well since the build tags are changing

Aceix (changes requested):

pro/initialize.go
Aceix: oh.. i now see the revert. might also want to check changes in .github/workflows/test.yml

gabrielseibel1: thx!

Aceix (approved):

mattkasun (approved):

abhishek9686 (approved):

PR #2558, opened 2023-09-06
fix(NET-581): add tenant tier to status api

Aceix:
## Describe your changes
allow ui to know tenant tier so branding can be applied be fore login

## Provide Issue ticket number if applicable/not in title
NET-581

- [ ] Netmaker is awesome.

### DIFF
--- a/controllers/server.go
+++ b/controllers/server.go
@@ -83,6 +83,7 @@ func getStatus(w http.ResponseWriter, r *http.Request) {
 		DB           bool   `json:"db_connected"`
 		Broker       bool   `json:"broker_connected"`
 		LicenseError string `json:"license_error"`
+		IsEE         string `json:"IsEE"`
 	}
 
 	licenseErr := ""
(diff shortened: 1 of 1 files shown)


pwilson77 (approved):

controllers/server.go
```diff
@@ -83,17 +83,24 @@ func getStatus(w http.ResponseWriter, r *http.Request) {
 		DB           bool   `json:"db_connected"`
 		Broker       bool   `json:"broker_connected"`
 		LicenseError string `json:"license_error"`
```
mattkasun, line 85: Can this be a bool? Then you could set it to servcfg.IsPro

Aceix: done

mattkasun (approved):

controllers/server.go
```diff
@@ -94,6 +95,7 @@ func getStatus(w http.ResponseWriter, r *http.Request) {
 		DB:           database.IsConnected(),
 		Broker:       mq.IsConnected(),
 		LicenseError: licenseErr,
+		IsEE:         servercfg.IsPro,
```
abhishek9686, line 98: Going forward, the plan is to use only `Pro` terminology instead of `EE`

abhishek9686 (changes requested):

controllers/server.go
Aceix: changed

abhishek9686 (approved):

PR #2606, opened 2023-09-26
go nm-install 

mattkasun:


mattkasun:
> nm-quick.sh was updated in a different PR. how can further env file changes be made if this PR is merged? i was under the assumption the installer code would be in this repo, with it's build and deploy workflow

the installer code is in /gravitl/nm-install

theguy951357 (changes requested):

PR #2616, opened 2023-10-08
servercfg.GetPublicIP fixes and unit tests

yunginnanet:
Bing Bong! It's me, ya boi.

## Describe your changes

- [Fix: errors in GetPublicIP ignored, err is always nil](https://github.com/gravitl/netmaker/commit/574e2ff058f6931bba9da2918562d89f2a652b3d)
- [Feat[testing]: Add unit tests for servercfg.GetPublicIP](https://github.com/gravitl/netmaker/commit/bf777152b22653992267c23342ef1ba49b6f11dc) + [Testing: Assert we fixed the bug in](https://github.com/gravitl/netmaker/pull/2616/commits/13978d79b73c1f75b4b977573e81aee515cfcce8) https://github.com/gravitl/netmaker/commit/574e2ff058f6931bba9da2918562d89f2a652b3d
- [Chore: gofmt](https://github.com/gravitl/netmaker/commit/6b60dc04fbe334ac924c6973bc3b1e191f0b32ff)

## Other Notes

recommend using github's squash merge, but i can squash it for you if you'd like.


yunginnanet:
lmao this PR sent me into a whole adventure involving an open pullrequest to golang/go that has one positive code review so far https://go-review.googlesource.com/c/go/+/533119

mattkasun (changes requested):
netclient now uses Stun to get public ip and thus this endpoint is deprecated and should be deleted (see NET-661)

yunginnanet:
hold on there partner

this function is heavily used within the configuration package

see servercfg

mattkasun:
> 🤠 hold on there partner 🤠 @mattkasun
> 
> this function is heavily used within the configuration package, not just that deprecated endpoint
> 
> see servercfg

my bad... mistakenly thought this PR was regarding controllers/getPublicIP() 
proceed

servercfg/serverconf.go
```diff
@@ -479,17 +479,36 @@ func GetPublicIP() (string, error) {
...
-		resp, err := client.Get(ipserver)
+		var resp *http.Response
+		resp, err = client.Get(ipserver)
 		if err != nil {
 			continue
 		}
 		defer resp.Body.Close()
 		if resp.StatusCode == http.StatusOK {
-			bodyBytes, err := io.ReadAll(resp.Body)
+			var bodyBytes []byte
```
mattkasun, line 510: no need to declare a var here  `bodyBytes, err :=io.ReadAll(resp.Body)`

mattkasun (changes requested):

servercfg/serverconf.go
yunginnanet: Well... this actually starts to touch on the bug that I'm fixing in the first place. The problem here is that we were checking for a non nil err afterwards that was... always nil... what I'm doing here is making sure that we are propagating that error value throughout the process so that by the time we check for a non-nil error, it will actually potentially contain a value 😅

mattkasun: the non-nil check after completing the loop is not required.  It does not provide any additional information.  Once the loop is completed, if the endpoint string is still empty, the error "public address not found" should be returned.

yunginnanet: okay, so wouldn't we like to know.. why? as my unit test shows, with this PR, we do. I'd like to know the reason that this function failed if it failed. are you saying we're not handling the error from this function? if so then that'd be what I'll attack next.

mattkasun: for the most part, this error is ignored by calling functions.  In the one case where it is checked, an err will  result in a fatalLog.

if we need the actual error for troubleshooting then just add a slog.Debug()  when the error occurs.

mattkasun (changes requested):
log the error if you think it is useful for troubleshooting

servercfg/serverconf.go
```diff
@@ -479,17 +479,36 @@ func GetPublicIP() (string, error) {
 		iplist = append([]string{publicIpService}, iplist...)
 	}
 
+	// regular timeout
+	timeoutTarget := time.Duration(10) * time.Second
+
+	// our unit test sets a bad ip service to test the timeout and fallback mechanics
+	// if that env variable is set, lower the timeout so the test doesn't take forever
+	testCheck := os.Getenv("NETMAKER_TEST_BAD_IP_SERVICE")
```
abhishek9686, line 487: tests shouldn't part of the main code logic

abhishek9686 (changes requested):

PR #2631, opened 2023-10-17
NET-674: clear extclients of an user when deleted, remove gw id from user gws when deleted

abhishek9686:
- [ ] Netmaker is awesome.

### DIFF
--- a/controllers/node.go
+++ b/controllers/node.go
@@ -571,10 +571,28 @@ func deleteIngressGateway(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if servercfg.IsPro && wasFailover {
-		if err = logic.EnterpriseResetFailoverFunc(node.Network); err != nil {
-			logger.Log(1, "failed to reset failover list during failover create", node.ID.String(), node.Network)
+	if servercfg.IsPro {
+		if wasFailover {
+			if err = logic.EnterpriseResetFailoverFunc(node.Network); err != nil {
+				logger.Log(1, "failed to reset failover list during failover create", node.ID.String(), node.Network)
+			}
 		}
+		go func() {
+			users, err := logic.GetUsersDB()
+			if err == nil {
+				for _, user := range users {
+					if _, ok := user.RemoteGwIDs[nodeid]; ok {
+						delete(user.RemoteGwIDs, nodeid)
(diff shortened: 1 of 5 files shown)


Aceix (approved):

controllers/user.go
```diff
@@ -525,7 +525,23 @@ func deleteUser(w http.ResponseWriter, r *http.Request) {
 		logic.ReturnErrorResponse(w, r, logic.FormatError(err, "badrequest"))
 		return
 	}
-
+	// check and delete extclient with this ownerID
+	go func() {
+		extclients, err := logic.GetAllExtClients()
+		if err == nil {
```
gabrielseibel1, line 531: could invert if and reduce indention.
like having
```go
err != nil { log and return }
dostuff...
```

logic/auth.go
```diff
@@ -39,6 +39,30 @@ func HasSuperAdmin() (bool, error) {
 	return false, err
 }
 
+// GetUsersDB - gets users
+func GetUsersDB() ([]models.User, error) {
```
gabrielseibel1, line 43: is this `func GetUsers() ([]models.ReturnUser, error) {` but with a different user type? could have a single function and an adapter from `models.User` to `models.ReturnUser` so we don't need to maintain duplicate code

controllers/node.go
```diff
@@ -571,10 +571,28 @@ func deleteIngressGateway(w http.ResponseWriter, r *http.Request) {
...
-	if servercfg.IsPro && wasFailover {
-		if err = logic.EnterpriseResetFailoverFunc(node.Network); err != nil {
-			logger.Log(1, "failed to reset failover list during failover create", node.ID.String(), node.Network)
+	if servercfg.IsPro {
+		if wasFailover {
+			if err = logic.EnterpriseResetFailoverFunc(node.Network); err != nil {
+				logger.Log(1, "failed to reset failover list during failover create", node.ID.String(), node.Network)
+			}
 		}
+		go func() {
```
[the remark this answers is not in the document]
abhishek9686: doesn't have to be, in non pro they cannot attach gateways to user anyways

controllers/user.go
abhishek9686: done

PR #2858, opened 2024-03-15
fix(NET-1009): fix client connectivity metrics data

Aceix:
- [ ] Netmaker is awesome.

### DIFF
--- a/pro/logic/metrics.go
+++ b/pro/logic/metrics.go
@@ -104,7 +104,6 @@ func MQUpdateMetrics(client mqtt.Client, msg mqtt.Message) {
 }
 
 func updateNodeMetrics(currentNode *models.Node, newMetrics *models.Metrics) {
-
 	oldMetrics, err := logic.GetMetrics(currentNode.ID.String())
 	if err != nil {
 		slog.Error("error finding old metrics for node", "id", currentNode.ID, "error", err)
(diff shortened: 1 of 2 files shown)


pro/logic/metrics.go
```diff
@@ -121,21 +120,16 @@ func updateNodeMetrics(currentNode *models.Node, newMetrics *models.Metrics) {
...
-					extMetric.Connected = true
-				}
-			}
-			extMetric.NodeName = attachedClients[i].ClientID
-			delete(newMetrics.Connectivity, attachedClients[i].PublicKey)
-			newMetrics.Connectivity[attachedClients[i].ClientID] = extMetric
+	for i := range attachedClients {
+		slog.Debug("[metrics] processing attached client", "client", attachedClients[i].ClientID, "public key", attachedClients[i].PublicKey)
+		clientMetric := newMetrics.Connectivity[attachedClients[i].PublicKey]
+		if clientMetric.TotalReceived > 0 && clientMetric.TotalSent > 0 {
```
abhishek9686, line 126: we set the connected field right, while metric collection on netclient, why are we setting here again by checking the traffic data here

abhishek9686 (changes requested):

pro/logic/metrics.go
Aceix: well, i also dont think it's necessary but i left it as it was old code.
will remove and test again

Aceix: resolved

abhishek9686 (approved):

PR #2864, opened 2024-03-17
Fix ACL, host and node swagger API definitions

jonathroth:
## Describe your changes
Fix get all hosts response swagger type (returned single instead of multiple hosts)
Remove body parameter from GET ACL request (it's a GET request, it shouldn't have a body)
Use ApiNode response in requests that return it (according to the actual code, the body is ApiNode[], the previous LegacyNode declaration was missing a few fields)
Redact net.Address field from API Iface - this field is a net.Address which doesn't have a JSON marshaller function - which caused it to be serialized into an object with a byte[] and string even though both fields are technically byte[], a generated C# client failed on deserializing this field due to type mismatch

- [ ] Netmaker is awesome.


Aceix (approved):

models/api_host.go
```diff
@@ -38,9 +45,12 @@ func (h *Host) ConvertNMHostToAPI() *ApiHost {
...
 	a.FirewallInUse = h.FirewallInUse
 	a.ID = h.ID.String()
-	a.Interfaces = h.Interfaces
+	a.Interfaces = make([]ApiIface, len(h.Interfaces))
 	for i := range a.Interfaces {
-		a.Interfaces[i].AddressString = a.Interfaces[i].Address.String()
+		a.Interfaces[i] = ApiIface{
+			Name:          h.Interfaces[i].Name,
+			AddressString: h.Interfaces[i].Address.String(),
+		}
```
abhishek9686, line 53: @Aceix will this require any changes on the UI side to be compatible

abhishek9686 (approved):

models/api_host.go
Aceix: nah.. it's fine. UI uses addressString only

abhishek9686:
Thank you for your contribution, @jonathroth.🏅

jonathroth:
@abhishek9686 Feel free to merge :)
I can't merge myself because I don't have write access

PR #2921, opened 2024-05-02
NET-1180: fix install issue for latest debian and centos distributions

yabinma:
- [ ] Netmaker is awesome.

### DIFF
--- a/scripts/nm-quick.sh
+++ b/scripts/nm-quick.sh
@@ -124,6 +124,7 @@ setup_netclient() {
 	chmod +x netclient
 	./netclient install
 	echo "Register token: $TOKEN"
+	sleep 2
 	netclient register -t $TOKEN
 
 	echo "waiting for netclient to become available"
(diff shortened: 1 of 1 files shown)


abhishek9686:
Add appropriate PR Title, and PR description

scripts/nm-quick.sh
```diff
@@ -582,7 +653,21 @@ install_netmaker() {
...
+	elif [ -f /etc/amazon-linux-release ]; then
+		docker compose up -d --force-recreate
+	elif [ -f /etc/centos-release ]; then
+		docker compose up -d --force-recreate
+	elif [ -f /etc/redhat-release ]; then
+		docker compose up -d --force-recreate
+	elif [ "$(cat /etc/*-release |grep suse |wc -l)" -gt 0 ]; then
+		docker compose up -d --force-recreate
+	else
+		docker-compose up -d --force-recreate
```
abhishek9686, line 669: why do we still need `docker-compose`?

yabinma: In the old code logic, there are the code for Alpine, ArchLinux and BSD,  they are not covered in the test. Because there is no cloud version for these OSs.  There is no public ip for local vm, the installation process could not complete.  The "docker-compose" is for these unchanged OSs.

The changes and test covered:
Ubuntu
Debian
Fedora
Centos
Redhat
Amazon Linux
Rocky Linux
AlmaLinux

SUSE Linux is tested, but it does not work. There is no x86_64 docker packages for SUSE, only os390 version supported.
@abhishek9686

yabinma: Good point.  Deleted suse and added the OS support list if the OS is not in the list.

abhishek9686 (approved):

PR #2958, opened 2024-05-31
NET-1268:seperate static port and static endpoint

yabinma:
- [ ] Netmaker is awesome.

### DIFF
--- a/models/host.go
+++ b/models/host.go
@@ -66,7 +66,8 @@ type Host struct {
 	EndpointIPv6        net.IP           `json:"endpointipv6"            yaml:"endpointipv6"`
 	IsDocker            bool             `json:"isdocker"                yaml:"isdocker"`
 	IsK8S               bool             `json:"isk8s"                   yaml:"isk8s"`
-	IsStatic            bool             `json:"isstatic"                yaml:"isstatic"`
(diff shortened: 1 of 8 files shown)


models/host.go
abhishek9686, line 1: Keep current IsStatic as it is, only add IsStaticPort, otherwise this will affect older versions

abhishek9686 (changes requested):

models/node.go
```diff
@@ -569,7 +569,7 @@ func (n *Node) Legacy(h *Host, s *ServerConfig, net *Network) *LegacyNode {
 	//l.FailoverNode = n.FailoverNode
 	//l.IngressGatewayRange = n.IngressGatewayRange
 	//l.IngressGatewayRange6 = n.IngressGatewayRange6
-	l.IsStatic = formatBool(h.IsStatic)
+	l.IsStatic = formatBool(h.IsStaticPort)
```
abhishek9686, line 572: why are we using `IsStaticPort` here to initialise `IsStatic`  ?

yabinma: It's missed in last change.  done.

abhishek9686 (approved):

PR #2978, opened 2024-06-14
NET-1288:add egress range check with netmaker network address

yabinma:
- [x] Netmaker is awesome.


controllers/node.go
```diff
@@ -458,6 +463,15 @@ func deleteEgressGateway(w http.ResponseWriter, r *http.Request) {
 		logic.ReturnErrorResponse(w, r, logic.FormatError(err, "badrequest"))
 		return
 	}
+	var gateway models.EgressGatewayRequest
+	gateway.NetID = params["network"]
+	gateway.Ranges = node.EgressGatewayRanges
+	err = logic.ValidateEgressRange(gateway)
```
abhishek9686, line 469: why do we need validate check on delete req?

abhishek9686 (changes requested):

controllers/node.go
yabinma: removed in new commit

abhishek9686 (approved):

PR #2987, opened 2024-06-24
NET-1226: Scalability Improvements

yabinma:
- [ ] Netmaker is awesome.

### DIFF
--- a/models/host.go
+++ b/models/host.go
@@ -116,8 +116,8 @@ const (
 	UpdateKeys HostMqAction = "UPDATE_KEYS"
 	// RequestPull - request a pull from a host
 	RequestPull HostMqAction = "REQ_PULL"
-	// JoinPull - request a pull from a host without restart
-	JoinPull HostMqAction = "JOIN_PULL"
+	// SignalPull - request a pull from a host without restart
+	SignalPull HostMqAction = "SIGNAL_PULL"
 	// UpdateMetrics - updates metrics data
 	UpdateMetrics HostMqAction = "UPDATE_METRICS"
 )
--- a/models/host.go
+++ b/models/host.go
@@ -116,6 +116,8 @@ const (
 	UpdateKeys HostMqAction = "UPDATE_KEYS"
 	// RequestPull - request a pull from a host
 	RequestPull HostMqAction = "REQ_PULL"
+	// JoinPull - request a pull from a host without restart
+	JoinPull HostMqAction = "JOIN_PULL"
 	// UpdateMetrics - updates metrics data
 	UpdateMetrics HostMqAction = "UPDATE_METRICS"
 )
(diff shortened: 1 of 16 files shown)


models/host.go
```diff
@@ -116,6 +116,8 @@ const (
 	UpdateKeys HostMqAction = "UPDATE_KEYS"
 	// RequestPull - request a pull from a host
 	RequestPull HostMqAction = "REQ_PULL"
+	// JoinPull - request a pull from a host without restart
+	JoinPull HostMqAction = "JOIN_PULL"
```
abhishek9686, line 120: Can we change the name to 'SIGNAL_PULL'

abhishek9686 (changes requested):

models/host.go
yabinma: no problem.  Added a commit to update the name.    netclient change added as well.

abhishek9686 (approved):

mq/mq.go
```diff
@@ -34,8 +34,8 @@ func setMqOptions(user, password string, opts *mqtt.ClientOptions) {
 	opts.SetAutoReconnect(true)
 	opts.SetConnectRetry(true)
 	opts.SetCleanSession(true)
-	opts.SetConnectRetryInterval(time.Second * 4)
-	opts.SetKeepAlive(time.Minute)
+	opts.SetConnectRetryInterval(time.Second * 3)
+	opts.SetKeepAlive(time.Second * 5)
```
abhishek9686, line 38: @yabinma do you think it's fine to change the `SetKeepAlive` back to a minute?
I am just thinking if this will cause too many ping requests in a given time at Scale on SaaS

mq/mq.go
```diff
@@ -96,9 +96,9 @@ func SetupMQTT(fatal bool) {
 	})
 	opts.SetConnectionLostHandler(func(c mqtt.Client, e error) {
 		slog.Warn("detected broker connection lost", "err", e.Error())
-		c.Disconnect(250)
-		slog.Info("re-initiating MQ connection")
-		SetupMQTT(false)
+		// c.Disconnect(250)
+		// slog.Info("re-initiating MQ connection")
+		// SetupMQTT(false)
```
abhishek9686, line 101: should we uncomment this piece of code?

abhishek9686 (changes requested):

mq/mq.go
yabinma: This is not a big issue as I can see. There is only one server connecting to mq, the traffic should not be a problem.
I will change it back to 1 minute in next commit.   Changing to 5 seconds does not help in the mq issue we met. Also in emqx settings, the server side time settings is matched with 1 minute.

mq/mq.go
yabinma: Yes,  added a commit to revert the settings back.

abhishek9686 (approved):

PR #3014, opened 2024-07-24
NET-1440: scale test changes

yabinma:
- [ ] Netmaker is awesome.


abhishek9686:
@yabinma tests are failing

yabinma:
> @yabinma tests are failing

UT cases was not updated. Done and test passed.

logic/ippool.go
```diff
@@ -0,0 +1,300 @@
...
+
+	"github.com/c-robinson/iplib"
+	"github.com/gravitl/netmaker/database"
+	"github.com/gravitl/netmaker/logger"
+	"github.com/gravitl/netmaker/models"
+	"golang.org/x/exp/slog"
+)
+
+var (
+	ipPool      map[string]PoolMap
```
abhishek9686, line 18: shouldn't this be intialised here `make(map[string]PoolMap)` ?

logic/networks.go
```diff
@@ -320,6 +382,12 @@ func UpdateNetwork(currentNetwork *models.Network, newNetwork *models.Network) (
 		}
 		newNetwork.SetNetworkLastModified()
 		err = database.Insert(newNetwork.NetID, string(data), database.NETWORKS_TABLE_NAME)
+		if err == nil {
+			if servercfg.CacheEnabled() {
+				storeNetworkInCache(newNetwork.NetID, *newNetwork)
+				deleteNetworkFromCache(currentNetwork.NetID)
```
abhishek9686, line 388: why are we deleting here after storing?

logic/nodes.go
```diff
@@ -112,17 +112,20 @@ func GetNetworkNodesMemory(allNodes []models.Node, network string) []models.Node
 // UpdateNodeCheckin - updates the checkin time of a node
 func UpdateNodeCheckin(node *models.Node) error {
 	node.SetLastCheckIn()
+	//If cache enabled, only Update the Check-in TS in cache
+	if servercfg.CacheEnabled() {
+		storeNodeInCache(*node)
+		return nil
```
abhishek9686, line 118: updates should be stored in DB as well, here it's storing in cache and returning

main.go
```diff
@@ -37,6 +37,9 @@ func main() {
 	servercfg.SetVersion(version)
 	fmt.Println(models.RetrieveLogo()) // print the logo
 	initialize()                       // initial db and acls
+	logic.SetIpPool()
+	defer logic.ClearIpPool()
+	defer logic.WriteMetricsCacheToDB()
```
abhishek9686, line 42: Why are we deferring writing metrics cache to DB?
This may cause inconsistency in the data when the server quits abruptly

mq/mq.go
```diff
@@ -96,15 +92,15 @@ func SetupMQTT(fatal bool) {
...
 		slog.Info("re-initiating MQ connection")
-		SetupMQTT(false)
+		//SetupMQTT(false)
 
 	})
 	mqclient = mqtt.NewClient(opts)
 	tperiod := time.Now().Add(10 * time.Second)
 	for {
-		if token := mqclient.Connect(); !token.WaitTimeout(MQ_TIMEOUT*time.Second) || token.Error() != nil {
+		if token := mqclient.Connect(); token.Wait() && token.Error() != nil {
```
abhishek9686, line 103: what is the default wait time?

pro/logic/metrics.go
```diff
@@ -29,21 +100,40 @@ func GetMetrics(nodeid string) (*models.Metrics, error) {
...
+	}
 	return &metrics, nil
 }
 
 // UpdateMetrics - updates the metrics of a given client
 func UpdateMetrics(nodeid string, metrics *models.Metrics) error {
+	//if cache is enabled, only save the metric data to cache
+	if servercfg.CacheEnabled() {
+		storeMetricsInCache(nodeid, *metrics)
+		return nil
```
abhishek9686, line 114: DB write is being skipped here

logic/telemetry.go
```diff
@@ -13,8 +13,11 @@ import (
 	"golang.org/x/exp/slog"
 )
 
-// flags to keep for telemetry
-var isFreeTier bool
+var (
+	// flags to keep for telemetry
+	isFreeTier      bool
+	telServerRecord = models.Telemetry{}
```
abhishek9686, line 19: telmetry record gets updated as well, this variable won't be in sync with DB

mq/publishers.go
```diff
@@ -134,6 +148,101 @@ func NodeUpdate(node *models.Node) error {
 	return nil
 }
 
+func ServerStatusUpdate() error {
```
abhishek9686, line 151: why do we need this functionality?
status API is not costly, having these operations over mq might be an overkill

abhishek9686 (changes requested):

main.go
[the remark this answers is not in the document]
yabinma: Change it back in new commit. 

Regarding Check-in and metric data, every time when updating check-in or metrics, it writes the data to DB, overwriting the old data.  The UI only load the latest data.  From performance perspective,  writing to memory is better than writing to DB.

In the 1000 hosts scale test,  DB write is a not a primary blocker issue.  I changed it back to avoid inconsistent issue.

logic/auth.go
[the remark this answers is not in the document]
yabinma: when superuser is updated, it is written to cache as well.  code changed in new commit.

logic/enrollmentkey.go
```diff
@@ -29,6 +31,10 @@ var EnrollmentErrors = struct {
 	FailedToTokenize:   fmt.Errorf("failed to tokenize"),
 	FailedToDeTokenize: fmt.Errorf("failed to detokenize"),
 }
+var (
+	enrollmentkeyCacheMutex = &sync.RWMutex{}
+	enrollmentkeyCacheMap   = make(map[string]*models.EnrollmentKey)
```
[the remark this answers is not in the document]
yabinma: done in new commit

logic/extpeers.go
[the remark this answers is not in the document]
yabinma: changed it back in new commit

logic/ippool.go
yabinma: done

logic/networks.go
yabinma: done

logic/nodes.go
yabinma: done

main.go
yabinma: done

mq/mq.go
yabinma: done

logic/telemetry.go
yabinma: done,  cache updated when telemetry is updated.

mq/publishers.go
yabinma: In mq approach, server only needs to send one message(one connection, rather than every client api call).
There is also another func to check if there is server status change. Only the server status is changed, a new mq message is distributed to clients.   It will save a lot of api requests.

mq/util.go
```diff
@@ -78,15 +78,16 @@ func publish(host *models.Host, dest string, msg []byte) error {
 	if encryptErr != nil {
 		return encryptErr
 	}
-	if mqclient == nil || !mqclient.IsConnectionOpen() {
+	if mqclient == nil || !mqclient.IsConnected() {
```
abhishek9686, line 81: won't this be more accurate check with `IsConnectionOpen` ?
otherwise this will cause timeout errors, if the connection is not active

abhishek9686 (changes requested):

mq/util.go
```diff
@@ -79,11 +78,11 @@ func publish(host *models.Host, dest string, msg []byte) error {
...
 		return errors.New("cannot publish ... mqclient not connected")
 	}
 
-	if token := mqclient.Publish(dest, 0, true, encrypted); !token.WaitTimeout(MQ_TIMEOUT*time.Second) || token.Error() != nil {
+	if token := mqclient.Publish(dest, 0, true, encrypted); token.Wait() && token.Error() != nil {
```
abhishek9686, line 85: `func (mqtt.Token) Wait() bool
Wait will wait indefinitely for the Token to complete, ie the Publish to be sent and confirmed receipt from the broker.`

we shouldn't wait forever, there should be a timeout, what's the thought process behind this?

logic/zombie.go
```diff
@@ -135,6 +135,8 @@ func ManageZombies(ctx context.Context, peerUpdate chan *models.Node) {
 					}
 				}
 			}
+			//reset the ip pool
+			SetIpPool()
```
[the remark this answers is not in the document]
yabinma: This is for the ip recycle.  If there is deleted node, the ip will be back to pool here.

mq/util.go
yabinma: Yes, IsConnectionOpen is the func to check if it's connected in deed.   IsConnected includes the scenarios if it's in re-connecting...
let me change it to IsConnectionOpen. But it will impact the publish as well. If it's in re-connecting status, the publish will be dropped.

logic/zombie.go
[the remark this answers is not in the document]
abhishek9686: so we don't add the ips back to the pool, just after the deletion of the node?

mq/mq.go
```diff
@@ -135,6 +131,9 @@ func Keepalive(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-time.After(time.Second * KEEPALIVE_TIMEOUT):
+			if mqclient == nil || !mqclient.IsConnectionOpen() {
+				SetupMQTT(false)
+			}
```
abhishek9686, line 136: this is cause duplicate client conns,  at times IsConnectionOpen might return false, but client can still be connected
I think we should get avoid performing setup conn again in the code

abhishek9686 (changes requested):

logic/zombie.go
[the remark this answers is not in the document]
yabinma: Adding ip back to pool in every node deletion is not an efficient way to recycle ip. 
Please help confirm to keep it is or change to recycle ip in every node delete.

mq/mq.go
yabinma: Removed the setupMQTT here. 
After the mq client close issue fixed, it's not necessary to re-establish connection.

logic/zombie.go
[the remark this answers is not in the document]
abhishek9686: > Adding ip back to pool in every node deletion is not an efficient way to recycle ip. Please help confirm to keep it is or change to recycle ip in every node delete.

we should add it back to pool after node deletion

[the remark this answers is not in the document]
yabinma: done

logic/ippool.go
abhishek9686, line 1: are we managing ip pool for extclients?

yabinma: Yes and No.
When the server startup and setIpPool() is called, the available ip calculates on both of netclient and extClient.

But when extClient is added, it does not pick up the ip from ip pool.  It was, but the change was reverted back based on discussion in https://github.com/gravitl/netmaker/pull/3014/commits/3d95e83f6eb8a1222024c87ac6a866a65d7fd487#r1694518855.

Usually, no impact to the ip allocation for netclient and extClient.    ip pool pop up ip with smallest number always; while extClient pick ip up in reverse direction. 
Only when the ip address is close to exhausted + the server never restart, there might be ip contention issue. 

Do you want me to add setIpPool() back to zombie cron? At least it will re-organize the ip pool regularly(for both of netclient and extClient).

logic/networks.go
```diff
@@ -15,13 +15,127 @@ import (
...
 )
 
+var (
+	networkCacheMutex = &sync.RWMutex{}
+	networkCacheMap   = make(map[string]models.Network)
+	allocatedIpMap    = make(map[string]map[string]net.IP)
+)
+
+const (
+	ipAddrCap = 5000
```
abhishek9686, line 30: do we need this ipcap?

logic/nodes.go
```diff
@@ -585,6 +593,14 @@ func createNode(node *models.Node) error {
 	if servercfg.CacheEnabled() {
```
abhishek9686, line 593: Add Ip to Map after node and extclient save to DB is successful

abhishek9686 (changes requested):

logic/networks.go
yabinma: change in new commit

logic/nodes.go
yabinma: done

logic/extpeers.go
```diff
@@ -90,6 +90,15 @@ func DeleteExtClient(network string, clientid string) error {
 	if err != nil {
 		return err
 	}
+	if extClient, ok := getExtClientFromCache(key); ok {
```
abhishek9686, line 93: if cache is disabled this will fail

logic/ippool.go
abhishek9686: Why are we still using ip pool?

abhishek9686 (changes requested):

logic/extpeers.go
yabinma: done

abhishek9686 (approved):

PR #3024, opened 2024-07-31
NET-1440:batch peerUpdate

yabinma:
- [ ] Netmaker is awesome.

### DIFF
--- a/mq/publishers.go
+++ b/mq/publishers.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/gravitl/netmaker/logger"
@@ -12,6 +13,8 @@ import (
 	"github.com/gravitl/netmaker/servercfg"
 	"golang.org/x/exp/slog"
 )
+
+const batchSize = 20
 
 // PublishPeerUpdate --- determines and publishes a peer update to all the hosts
 func PublishPeerUpdate(replacePeers bool) error {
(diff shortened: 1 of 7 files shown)


mq/publishers.go
abhishek9686, line 1: @yabinma we have an internal func to batch items, you may want to re-use it
```// BatchItems creates batches of up to size batchSize based on provided slice items
func BatchItems[T any](items []T, batchSize int) [][]T {
	if batchSize <= 0 {
		return nil
	}
	remainderBatchSize := len(items) % batchSize
	nBatches := int(math.Ceil(float64(len(items)) / float64(batchSize)))
	batches := make([][]T, nBatches)
	for i := range batches {
		if i == nBatches-1 && remainderBatchSize > 0 {
			batches[i] = make([]T, remainderBatchSize)
		} else {
			batches[i] = make([]T, batchSize)
		}
		for j := range batches[i] {
			batches[i][j] = items[i*batchSize+j]
		}
	}
	return batches
}```

File: mq/publishers.go [hunk 1]
```diff
@@ -13,6 +14,8 @@ import (
 	"golang.org/x/exp/slog"
 )
 
+const batchSize = 20
```
abhishek9686, line 17: should be read from config, we should also have a way to override the batch updates

abhishek9686 (changes requested):

mq/publishers.go
yabinma: Is this in netmaker repo?  I did not find it when searching with BatchItems  or batchSize.

mq/publishers.go
yabinma: It's good to set a switch to turn on/off batch peerUpdate. 
Regarding the batchsize,  it's necessary to open it to end user?  Please help confirm.

mq/publishers.go
abhishek9686: no this is in the saas repo

mq/publishers.go
abhishek9686: yes, allowing it to be configured will be flexible for us

yabinma: Added the peerUpdate batch size to config. Please help review.

mq/publishers.go
yabinma: done

mq/publishers.go
```diff
@@ -13,6 +14,8 @@ import (
 	"golang.org/x/exp/slog"
 )
 
+var batch int
+
```
abhishek9686, line 18: why is this var declared here?

abhishek9686 (changes requested):

mq/publishers.go
yabinma: > why is this var declared here?

For reading the env once, not reading the env every time when the func is called.

abhishek9686: var batch = servercfg.GetPeerUpdateBatchSize()

abhishek9686: we should also have a config to turn off batchupdates

yabinma: > var batch = servercfg.GetPeerUpdateBatchSize()

this is to call the env multiple times.  Anyway,  I will make change in next commit.

servercfg/serverconf.go
abhishek9686, line 1: Shouldn't the batch update be turned on by default?

abhishek9686 (changes requested):

servercfg/serverconf.go
yabinma: Thanks for the comment.    I have no idea which is better by default. 
Change done.

abhishek9686 (approved):

PR #3096, opened 2024-09-03
Updated to go1.23

lmnzx:
- [ ] Netmaker is awesome.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version-file: go.mod
+          go-version-file: 'go.mod'
       - name: Build
         run: |
          env CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build main.go
@@ -29,7 +29,7 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v5
         with:
-          go-version-file: go.mod
+          go-version-file: 'go.mod'
       - name: Build
         run: |
           cd cli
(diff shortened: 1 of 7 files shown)


abhishek9686 (approved):

.github/workflows/test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.19
+          go-version-file: go.mod
```
abhishek9686, line 32: Are quotes not required around go.mod?

lmnzx: they are optional, i will still add them as to maintain the code style

abhishek9686 (approved):

docker/Dockerfile-go-builder
abhishek9686, line 1: we should follow versioned tag here https://github.com/gravitl/netmaker/blob/76027a6115103ea38258632aa26f72fe5ba1d09b/Dockerfile#L2

we can push go builder image with tag 1.23.0

lmnzx: sorry, i over looked it, fixing it

abhishek9686 (approved):

docker/Dockerfile-go-builder
abhishek9686: https://github.com/gravitl/netmaker/blob/76027a6115103ea38258632aa26f72fe5ba1d09b/.github/workflows/docker-builder.yml#L29 tag in this workflow is pointing to latest, also seems like this is running in push to develop always which is unnecessary

abhishek9686 (approved):

PR #3107, opened 2024-09-06
NET-1509:add ingresspersistentkeepalive and ingressmtu for extClient/RAC config

yabinma:
- [ ] Netmaker is awesome.

### DIFF
--- a/models/node.go
+++ b/models/node.go
@@ -77,16 +77,18 @@ type CommonNode struct {
 // Node - a model of a network node
 type Node struct {
 	CommonNode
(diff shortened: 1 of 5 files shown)


models/node.go
abhishek9686, line 1: why do we need additional properties here, host model  already has MTU and PKA?

abhishek9686 (changes requested):

models/node.go
yabinma: The ones in host model are for host.
The  MTU and PKA in the ticket are for extClient/RAC only(no change for host/RAGW).

If re-using the ones in host,  when updating the values, it will impact RAGW itself.

abhishek9686 (approved):

PR #3124, opened 2024-09-13
NET-1603: Manage DNS NM changes

yabinma:
- [ ] Netmaker is awesome.


mq/mq.go
abhishek9686, line 1: remove periodic sync check, instead this can be published when peer update is pushed

mq/publishers.go
```diff
@@ -249,3 +250,48 @@ func sendPeers() {
 		}
 	}
 }
+
+func sendDNSSync() error {
```
abhishek9686, line 254: this func should be simplified to all get all entries from the DB and push it to clients

abhishek9686 (changes requested):

mq/mq.go
yabinma: Move to peerUpdate in new commit. please review again.

mq/publishers.go
yabinma: Removed the cache in new commit. Please help review again.

abhishek9686:
add relevant PR title

abhishek9686 (approved):

PR #3512, opened 2025-06-17
fix: update extclient ingress endpoint/port with host changes

Aceix:
- [ ] Netmaker is awesome.


controllers/hosts.go
abhishek9686, line 1: This logic is unnecessary, we already get latest data here in this api https://github.com/gravitl/netmaker/blob/9bd20dcec5f38347585050cb7c48f667eeda4034/pro/controllers/users.go#L63

PR #3576, opened 2025-07-29
NM-38: User Config Fixes

VishalDalwadi:
- [ ] Netmaker is awesome.


pro/controllers/users.go
```diff
@@ -1279,58 +1280,102 @@ func getUserRemoteAccessGwsV1(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	userGwNodes := proLogic.GetUserRAGNodes(*user)
+
```
abhishek9686, line 1283: Should check this logic
I believe this change was causing issues, so we had to revert the change from the release patch.

Mejdii (approved):

PR #3729, opened 2025-11-08
controllers: return 404 on get network not found

clwluvw:
## Describe your changes

return http 404 when network is not there rather than an internal err.

## Provide testing steps

Call `GET /api/networks/{networkname}` on a non-existent network to get a 404 rather than a 500 HTTP status code.

- [x] Netmaker is awesome.


abhishek9686 (approved):
Thanks for your contribution 🏅

